PR #22, opened 2016-12-14
build(ls): move to using yarn and yarn.lock files

chuckjaz:
yarn.lock at root is a signal to travis to install yarn.


vicb:
LGTM

PR #54, opened 2017-03-27
fix: improve error reporting on server requests

chuckjaz:


tbosch:
Is there a way to test this?

PR #138, opened 2017-08-03
Added error handling for @angular/language-service calls

### LINKED ISSUE #128: Output Window gets opened when I type
Output window gets opened any time I type some code and shows language service output.

When I close it doesn't have permanent effect, it gets opened again when I type.

I can't see any settings in the plugin. What shall I do? I don't want Output window open.

DBosley:
This PR should fix #128 for any errors thrown by the following @angular/language-service calls:
```
getHoverAt
getDefinitionAt
getTemplateReferences
getDiagnostics
getCompletionsAt
```


server/src/editorServices.ts
```diff
@@ -3035,7 +3035,15 @@ export class LineLeaf implements LineCollection {
...
   function time<T>(name: string, cb: () => T): T {
     const start = Date.now();
-    const result = cb();
+    let result: T = null;
+    try{
+        result = cb();
+    } catch (error) {
+        logger.msg(
+            `Error for ${name}:\n` +
+            `    ${error}`
```
chuckjaz: The stack is useful to diagnose problems. Consider changing this to `error.stack || error`

chuckjaz:
Let me know if you want to take my recommendation. I will commit this before the release as well.

DBosley:
@chuckjaz I'll toss that in there. One minute.

DBosley:
@chuckjaz All set!

Issue #343, opened 2019-07-04
Extension does not work after updating to Angular v8.1.0


ramunas-k:
I use this extension a lot, but today after updating to Angular v8.1.0 I noticed that I'm not getting any support from it. Here is the extension output:
```
Search path: /home/user/git/project/src/app/campaigns/campaign-list
Config file name: /home/user/git/project/tsconfig.json
Angular Language Service: 8.1.0
TypeScript: 3.4.5
SERVER ERROR: this.ngHost.setSite is not a function
TypeError: this.ngHost.setSite is not a function
    at new CompilerService (/home/user/.vscode/extensions/angular.ng-template-0.800.0/server/editorServices.js:1754:21)
    at new Project (/home/user/.vscode/extensions/angular.ng-template-0.800.0/server/editorServices.js:705:36)
    at ProjectService.createProject (/home/user/.vscode/extensions/angular.ng-template-0.800.0/server/editorServices.js:1726:23)
    at ProjectService.openConfigFile (/home/user/.vscode/extensions/angular.ng-template-0.800.0/server/editorServices.js:1629:28)
    at ProjectService.openOrUpdateConfiguredProjectForFile (/home/user/.vscode/extensions/angular.ng-template-0.800.0/server/editorServices.js:1489:41)
    at ProjectService.openClientFile (/home/user/.vscode/extensions/angular.ng-template-0.800.0/server/editorServices.js:1407:23)
    at /home/user/.vscode/extensions/angular.ng-template-0.800.0/server/documents.js:135:47
    at TextDocuments.logErrors (/home/user/.vscode/extensions/angular.ng-template-0.800.0/server/documents.js:235:13)
    at /home/user/.vscode/extensions/angular.ng-template-0.800.0/server/documents.js:131:74
    at handleNotification (/home/user/.vscode/extensions/angular.ng-template-0.800.0/server/node_modules/vscode-jsonrpc/lib/main.js:489:43)
[Error - 10:41:05 AM] Notification handler 'textDocument/didOpen' failed with message: this.ngHost.setSite is not a function
```
`ng version` output:
```
Angular CLI: 8.1.0
Node: 12.6.0
OS: linux x64
Angular: 8.1.0
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.801.0
@angular-devkit/build-angular     0.801.0
[...]

PR #344, opened 2019-07-04
Upgrade to Angular 8.1.0

andrius-pra:
Fixes: #343


andrius-pra:
I signed it!

david0178418:
Holding my breath for this change to be merged and released.

kyliau:
Thank you for fixing this! It's totally my oversight for failing to take into account the usage here when making the breaking change.

PR #349, opened 2019-08-08
feat: Backwards compat with new TS LS interface

kyliau:
https://github.com/angular/angular/pull/31972 changes the Angular LS
interface to return TS data types. This PR makes the extension
compatible with both the old and new interface.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -5,12 +5,8 @@
  * ------------------------------------------------------------------------------------------ */
 
 import * as ng from '@angular/language-service';
-
-import {
-  createConnection, IConnection, InitializeResult, TextDocumentPositionParams,
-  CompletionItem, CompletionItemKind, Definition, Location, TextDocumentIdentifier,
-  Position, Range, TextEdit, Hover
-} from 'vscode-languageserver';
+import * as ts from 'typescript';
+import * as lsp from 'vscode-languageserver';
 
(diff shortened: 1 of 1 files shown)


server/src/server.ts
```diff
@@ -165,22 +162,51 @@ function logErrors<T>(f: () => T): T {
...
+    const {textSpan, definitions} = result as ts.DefinitionInfoAndBoundSpan;
+    if (!definitions || !definitions.length) {
+      return;
+    }
+    const [start, end] = documents.offsetsToPositions(textDocument, [
+      textSpan.start,
+      textSpan.start + textSpan.length,
+    ]);
+    const originSelectionRange = lsp.Range.create(
+      start.line - 1, start.character - 1, end.line - 1, end.character - 1);
```
andrius-pra: @kyliau, you don't need to adjust line and character position. It already have correct position.

kyliau: Isn't vscode 1-based whereas LSP is 0-based?

andrius-pra: I don't know. I only tested changes from angular/angular#32098 and angular/angular#32085. I saw these bugs: 
"Go To Definition" doesn't work for html tags.
"Go To Definition" navigates to wrong line and position for html attributes.
 It work as expected when I remove " -1" from 186 and 194 lines

kyliau: thanks, I'll double check this.

PR #352, opened 2019-08-11
share DocumentRegistry instance between all language service instances

andrius-pra:
This PR makes the extension to share [DocumentRegistry](https://github.com/microsoft/TypeScript/blob/e46c846ee6e20687ea4a39055f1081ba73ca89ec/src/services/documentRegistry.ts#L3-L17) instance between all instances of LanguageService.

### DIFF
--- a/server/src/editorServices.ts
+++ b/server/src/editorServices.ts
@@ -802,6 +802,7 @@ export class Project {
   constructor(
       public projectService: ProjectService,
       private logger: Logger,
+      private documentRegistry: ts.DocumentRegistry,
       public projectOptions?: ProjectOptions,
       public languageServiceDisabled = false) {
       if (projectOptions && projectOptions.files) {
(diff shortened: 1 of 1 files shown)


kyliau:
@andrius-pra Thank you for all your contributions!
After talking to the the TypeScript team, we think it's best to stick with LSP instead of trying to make tsserver work with external files (HTML).
There are many reasons to this, but the main one is compatibility with other editors, not just vscode. LSP allows us to maintain full control of TypeScript server.
I made a prototype that completely eliminates managing the DocumentRegistry on our own, and instead delegate that to tsserver instead.
You might be interested: https://github.com/kyliau/vscode-ng-language-service/blob/prototype/server/src/server.ts
Please let me know your thoughts! And again, thank you very much =)

server/src/editorServices.ts
```diff
@@ -1208,7 +1210,7 @@ export class ProjectService {
   }
 
   createInferredProject(root: ScriptInfo) {
-      const project = new Project(this, this.psLogger);
+      const project = new Project(this, this.psLogger, this.documentRegistry, undefined, false);
```
kyliau: Consider leaving out the last two parameters since there are default values.

andrius-pra:
Your prototype looks great! Do we still need changes from typescript side to properly support external files? Your prototype only work in external templates with typescript [fork](https://github.com/microsoft/TypeScript/compare/master...andrius-pra:plugin-external-files) that already has changes for external template support in tsserver.

kyliau:
In the prototype, because we control the entire TS server, there is no change needed in TypeScript to support external template. The fork is still needed if we want tsserver **plugin** itself to support external template.

kyliau:
By the way, @andrius-pra are you on Slack? Would love to add you to the Angular Language Service channel. Please ping me on Twitter `@liauky` :)
We appreciate all the work that you've done to improve the language service!

andrius-pra:
@kyliau, I don't use Twitter, but you can find me in a [gitter](https://gitter.im).

PR #362, opened 2019-08-23
feature: Add components for the new prototype

kyliau:
This commit adds the helper classes to the main repo in preparation for the migration to the new codebase.

Context: Previously, `editorServices.ts` was cloned from TypeScript and modified to handle HTML files. This incurs maintenance liability and performance cost. The new architecture imports TypeScript `ProjectService` and wraps it so that HTML files are handled correctly.


server/src/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+enum Scheme {
+  File = 'file',
+}
+;
+
+/**
+ * Extract the file path from the specified `uri`.
+ * @param uri
+ */
+export function uriToFilePath(uri: string): string {
```
andrius-pra: This method returns wrong value for windows file names. for `file:///e%3A/demo/src/app/app.component.html` returns `/e:/demo/src/app/app.component.html` instead of `e:/demo/src/app/app.component.html`. It's already solved the fix is [here](https://github.com/angular/vscode-ng-language-service/blob/master/server/src/documents.ts#L109-L111)

kyliau: Could you please double check that this is an actual bug? Does tsserver handle the path correctly even with a leading `/`?
I'm sorry I don't have a Windows machine available so I can't test this. In any case, we should file an issue with upstream `vscode-uri` if this is a bug.
https://github.com/microsoft/vscode-uri

andrius-pra: I have tested in your prototype branch. It doesn't work from windows.
I get this error: 'No config file for /e:/demo/ls-demo/src/app/app.component.ts'.

it starts to work when I change `path` to `fsPath`.
image

kyliau: Thank you for the confirmation! I'll fix it.

server/src/completion.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Convert ts.CompletionEntry to LSP Completion Item.
+ * @param entry completion entry
+ * @param position position where completion is requested.
+ */
+export function tsCompletionEntryToLspCompletionItem(
+    entry: ts.CompletionEntry, position: lsp.Position): lsp.CompletionItem {
+  const item = lsp.CompletionItem.create(entry.name);
+  // TODO: Even though `entry.kind` is typed as ts.ScriptElementKind, it's
```
ayazhafiz: Can we type it is a union or always assign `ts.ScriptElementKind` values to begin with? This seems dangerous.

server/src/logger.ts
```diff
@@ -0,0 +1,153 @@
...
+import * as ts from 'typescript/lib/tsserverlibrary'; // used as value
+
+// NOTES:
+// Be very careful about logging. There are two types of logging:
+// 1. Console
+// 2. File
+// The language server could operate in a few different modes, depending on
+// startup options. By default, it communicates with the client through
+// Node IPC. In this case, logging to stdout/stderr would be piped to VSCode's
+// development console. Verbose log should not be sent to the client as it
```
ayazhafiz: Is the dev console sent to the client?

server/src/project_service.ts
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * `ProjectService` is a singleton service for the entire lifespan of the
+ * language server. This specific implementation is a very thin wrapper
+ * around TypeScript's `ProjectService`. On creation, it spins up tsserver and
+ * loads `@angular/language-service` as a tsserver plugin.
+ * Care must be taken to load `@angular/language-service` from the project or
+ * workspace. This extension shoud not bundle its own
+ * `@angular/language-service`, not even as a backup. This is to ensure that
+ * the extension becomes a no-op when non-Angular TS projects are opened.
```
ayazhafiz: I thought in the case that an only-TS project is opened, the language service still runs but _always_ defers to the TypeScript language service. Is this not correct?

Can we guarantee that Angular users will have their own `@angular/language-service`? Do they have to specify it as a dependency for the plugin to work?

server/src/project_service.ts
```diff
@@ -0,0 +1,142 @@
...
+      eventHandler: (e) => this.handleProjectServiceEvent(e),
+      globalPlugins: ['@angular/language-service'],
+      pluginProbeLocations: [pluginProbeLocation],
+      allowLocalPluginLoads: false,  // do not load plugins from tsconfig.json
+    });
+
+    this.tsProjSvc.configurePlugin({
+      pluginName: '@angular/language-service',
+      configuration: {
+        'angularOnly': true,
```
ayazhafiz: This still provides TypeScript functionality, right?

server/src/project_service.ts
```diff
@@ -0,0 +1,142 @@
...
+   * make that project its default. This method is to ensure HTML files always
+   * belong to a configured project instead of the default behavior of being in
+   * an inferred project.
+   * @param scriptInfo
+   */
+  getDefaultProjectForScriptInfo(scriptInfo: ts.server.ScriptInfo): ts.server.Project
+      |undefined {
+    let project = this.tsProjSvc.getDefaultProjectForFile(
+        scriptInfo.fileName,
+        false  // ensureProject TODO: Document what this means
```
ayazhafiz: Is it costly to document what this means now?

server/src/server_host.ts
```diff
@@ -0,0 +1,173 @@
+/**
+ * `ServerHost` is a wrapper around `ts.sys` for the Node system. In Node, all
+ * optional methods of `ts.System` are implemented.
```
ayazhafiz: I'm a bit confused by this. How do we know that the `ts.System` passed to the constructor will be the system created from `sys.getNodeSystem`?

server/src/utils.ts
```diff
@@ -0,0 +1,75 @@
+import * as lsp from 'vscode-languageserver';
+import {URI} from 'vscode-uri';
+
+enum Scheme {
+  File = 'file',
+}
+;
```
ayazhafiz:
```suggestion
```

server/src/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Extract the file path from the specified `uri`.
+ * @param uri
+ */
+export function uriToFilePath(uri: string): string {
```
ayazhafiz: Will you get rid of this method when you fix the conversation above?

server/src/logger.ts
kyliau: The dev console is the "Output" tab in VSCode, so, yeah, it is sent to client

ayazhafiz: Can you clarify this in the comments?

server/src/project_service.ts
kyliau: The first statement is incorrect. The goal of the Angular LS is to answer Angular queries only. That's why the `angularOnly` flag was introduced in the plugin.

Our extension loads `@angular/language-service` as a global plugin.

Users who want to use the Angular LS must install `@angular/language-service` in their project (it's installed by default if users use the Angular CLI). If the extension does not find the package in the workspace, it should bail, effectively becoming a no-op extension.

That said, if `@angular/language-service` is loaded as a **tsserver local plugin**, it should never disable TypeScript language service. Does that make sense?

server/src/project_service.ts
kyliau: No, it does not provide TypeScript functionality. Please see comment above.

server/src/utils.ts
kyliau: ? this method is still needed.
Perhaps it's less confusing if I rename it to `uriToFsPath` ?

ayazhafiz: Can't you just use the function from https://github.com/angular/vscode-ng-language-service/blob/master/server/src/documents.ts#L109-L111? I'm probably missing something, sorry 😅

server/src/project_service.ts
ayazhafiz: When would the Angular language service be loaded as a global plugin versus a local plugin? If it is a global plugin and angularOnly, would this mean that TypeScript language service would not be enabled?

You’ve explained this to me before, but I think my mental model has degraded; sorry about that :)

server/src/utils.ts
kyliau: ahh ... sorry! I should have clarified, in a follow-up PR, the following files will be deleted.
1. `documents.ts`
2. `editorServices.ts`
3. `errors.ts`

and `server.ts` would be totally rewritten. The files in this PR is for the new `server.ts`.

We should never implement custom logic for URI parsing, instead should always call methods in `vscode-uri`.

server/src/project_service.ts
kyliau: Instead of explaining the difference between local plugin and global plugin here, I'll write a doc in the wiki and capture the previous discussion. I started a page already, will add more entries to it later.

server/src/server_host.ts
kyliau: Sorry for the confusion. This is meant to be used by the new `server.ts`. But you're right, instead of accepting `ts.sys` as an input, it should just import it directly.

If TypeScript detects that it's running on Node, it'd create `sys = getNodeSystem()`.
https://github.com/microsoft/TypeScript/blob/25f609b3a22996ba576ff6d6254567a36cc274b6/src/compiler/sys.ts#L1382-L1386

server/src/utils.ts
ayazhafiz: Cool, thanks for the clarification.

PR #382, opened 2019-09-04
fix(server): don't skip definitions with a zero-length range

ayazhafiz:
Certain definition files (like CSS files) don't have a `scriptInfo` but
also don't need to be read, because their definition spans are of 0
length. Don't skip over such definitions.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -168,11 +168,18 @@ connection.onDefinition((params: lsp.TextDocumentPositionParams) => {
   const results: lsp.LocationLink[] = [];
   for (const d of definition.definitions) {
     const scriptInfo = tsProjSvc.getScriptInfo(d.fileName);
-    if (!scriptInfo) {
+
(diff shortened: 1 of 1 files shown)


server/src/server.ts
```diff
@@ -168,11 +168,18 @@ connection.onDefinition((params: lsp.TextDocumentPositionParams) => {
...
-    if (!scriptInfo) {
+
+    // Some definitions, like definitions of CSS files, may not be recorded files with a
+    // `scriptInfo` but are still valid definitions because they are files that exist. In this case,
+    // check to make sure that the text span of the definition is zero so that the file doesn't have
+    // to be read; if the span is non-zero, we can't do anything with this definition.
+    if (!scriptInfo && d.textSpan.length > 0) {
       continue;
     }
+    const ZERO_RANGE = lsp.Range.create(0, 0, 0, 0);  // for files without `scriptInfo`
```
kyliau: Please move constant to top level. Perhaps `EMPTY_RANGE` is a better name?

Issue #369, opened 2019-08-28
Extension failed to load angular plugin from nested folder.


andrius-pra:
The extension(v0.900.0-next.0) failed to load @angular/language-service plugin when angular project is in a child folder. It only tries to load from root of workspace.

Steps to reproduce:
```
mkdir root
cd root
ng new test-ng-ls --routing=true --style=less
cd test-ng-ls
ng update @angular/core@9.0.0-next.2 @angular/cli@9.0.0-next.1 codelyzer
code ..
```

Environment:
```
@angular/language-service@~9.0.0-next.2
angular.ng-template: v0.900.0-next.0
```

jpike88:
@kyliau having the exact same problem here. It's preventing me from using the plugin at all

I have a workspace structure like so:

```
workspace root -  
                        - client (contains tsconfig.json, angular.json etc)
                        - server (contains just a tsconfig.json)
                        - shared (just .ts files)
```

The plugin doesn't recognize my client folder.

Installing @angular/language-service in the root folder results predictably in errors across all my components (```Component '...' is not included in a module and will not be available inside a template```)... so that isn't a valid workaround.

jpike88:
@kyliau I should also add, I have a tsconfig.json in my root directory.

From inspecting your plugin's code, there appears to be a variable called ```pluginProbeLocation```, but it never gets set.

https://github.com/angular/vscode-ng-language-service/blob/acc2a70a80efa16ab876e38bfd8de508afb8915a/server/src/project_service.ts#L45

So the solution maybe is that some sort of override needs to be set, in order to ensure that ```@angular/language-service``` gets loaded from the correct subdirectory.

Can you please shed some light on this? I'd imagine quite a few developers use a workspace where the angular project is in a subdirectory.

jpike88:
@kyliau can I get an an idea of whether this issue has any level of priority? If not, I'll attempt a fix myself when I get the time. However, I don't want to if you can conceive of something much faster, as this plugin stuff is a little alien to me.

andrius-pra:
I installed `@angular/language-service` in a workspace root and it was enough to fix this issue.
```
mkdir root
cd root
yarn add --dev @angular/language-service@next // <-- Added this command
ng new test-ng-ls --routing=true --style=less
cd test-ng-ls
ng update @angular/core@next @angular/cli@next codelyzer
code ..
```

jpike88:
@andrius-pra if I do that, all my components get this error:

[...]

Issue #400, opened 2019-09-18
Extension not working and crashing


RobertDelmatic:
The extension does not work at all, for me, in HTML templates, and it keeps crashing. Version 0.802.3
This is the console log:

Search path: c:/work/clientportal/angular/src/app/project/dialogs/project-edit-user-dialog
Config file name: c:/work/clientportal/angular/tsconfig.json
Angular Language Service: 6.0.3
TypeScript: 3.5.3
getTemplateReferences: 5629ms
Add recursive watcher for: c:/work/clientportal/angular
Opened configuration file c:/work/clientportal/angular/tsconfig.json
getTemplateReferences: 0ms
getDiagnostics: 124ms
internal/process/next_tick.js:63
          Reflect.apply(callback, undefined, tock.args);
                  ^

TypeError: Reflect.apply is not a function
    at process._tickCallback (internal/process/next_tick.js:63:19)
[Info  - 16:15:39] Connection to server got closed. Server will restart.
[Error - 16:15:39] Request textDocument/hover failed.
Error: Connection got disposed.
    at Object.dispose (c:\Users\RobertM\.vscode\extensions\angular.ng-template-0.802.3\client\node_modules\vscode-jsonrpc\lib\main.js:876:25)
    at Object.dispose (c:\Users\RobertM\.vscode\extensions\angular.ng-template-0.802.3\client\node_modules\vscode-languageclient\lib\client.js:71:35)
    at LanguageClient.handleConnectionClosed (c:\Users\RobertM\.vscode\extensions\angular.ng-template-0.802.3\client\node_modules\vscode-languageclient\lib\client.js:2153:42)
    at LanguageClient.handleConnectionClosed (c:\Users\RobertM\.vscode\extensions\angular.ng-template-0.802.3\client\node_modules\vscode-languageclient\lib\main.js:151:15)
    at closeHandler (c:\Users\RobertM\.vscode\extensions\angular.ng-template-0.802.3\client\node_modules\vscode-languageclient\lib\client.js:2140:18)
    at CallbackList.invoke (c:\Users\RobertM\.vscode\extensions\angular.ng-template-0.802.3\client\node_modules\vscode-jsonrpc\lib\events.js:62:39)
    at Emitter.fire (c:\Users\RobertM\.vscode\extensions\angular.ng-template-0.802.3\client\node_modules\vscode-jsonrpc\lib\events.js:120:36)
[...]

PR #402, opened 2019-09-19
feat: Bundle @angular/language-service with server

kyliau:
The initial idea of the new extension is to always load
`@angular/language-service` from an Angular project that has the package
installed. That is why it does not bundle the package with the server.

However, `@angular/language-service` is not always available, due to a
few reasons:
1. The initial directory opened is not the root of an Angular project.
2. When spawned as pure LSP server, the process CWD is not where
@angular/langauge-service is installed.

Besides that, users have no control over which @angular/language-service
version they want the extension to use. (TypeScript extension provides
this feature).

It is also surprising to users that the extension does nothing if
@angular/language-service is not found.

For the reasons above, it is better that the server bundles a fallback
version of @angular/language-service, but still provides users with an
option to specify the location via `--pluginProbeLocation` cmd line
flag.

PR closes https://github.com/angular/vscode-ng-language-service/issues/369

### DIFF
--- a/server/src/project_service.ts
+++ b/server/src/project_service.ts
@@ -12,6 +12,7 @@ import * as lsp from 'vscode-languageserver';
 import {tsDiagnosticToLspDiagnostic} from './diagnostic';
 import {projectLoadingNotification} from './protocol';
 import {filePathToUri} from './utils';
+import {NgVersionProvider} from './version_provider';
 
 // NOTE:
 // There are three types of `project`:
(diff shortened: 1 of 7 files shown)


server/src/project_service.ts
```diff
@@ -172,4 +171,33 @@ export class ProjectService {
...
+  /**
+   * Determine valid directories where `@angular/language-service` could be
+   * found.
+   * @param probeLocation
+   */
+  private getPluginProbeLocations(probeLocation?: string): string[] {
+    const versionProvider = new NgVersionProvider(probeLocation);
+    const pluginProbeLocations = [];
+    const localVersion = versionProvider.localVersion;
+    if (localVersion) {
```
andrius-pra: Does the extension work with 8.x.x or bellow version? Consider to ignore unsupported versions. This will prevent the issues like #400.

kyliau: yes, I agree with that. Will do so in a follow up PR.

kyliau:
> This LGTM, but I think we should consider another name for the `pluginProbeLocation` CLI argument, which may be unintuitive for users. A user may more easily understand the meaning of something like `angularLSPath` , for example.

I was thinking of maintaining compatibility with the tsserver, hence the name for the flag. I'll create an issue to clean up the CLI options (along with help menu) before we make a final release.

PR #403, opened 2019-10-08
build: Add rollup to build pipeline

kyliau:
Rollup requires es6 (es2015) modules, so TypeScript output was changed to `es2015`. Final rollup output is `amd` so that we could provide custom impl for `define()`.


kyliau:
work in progress: es2015 mode breaks the test, will need to install additional rollup plugin.

kyliau:
PR is good to go, PTAL.

package.json
```diff
@@ -31,11 +31,11 @@
   ],
   "main": "./client/out/extension",
   "scripts": {
-    "compile": "tsc -b",
+    "compile": "tsc -b && rollup -c",
     "compile:test": "tsc -b server/src/tests",
     "compile:integration": "tsc -b integration",
     "format": "scripts/format.sh",
-    "watch": "tsc -b -w",
+    "watch": "tsc -b -w; rollup -c -w",
```
ayazhafiz: Maybe use `&` instead of `;` if we want to launch both of these processes, as `tsc -b -w` will block?

server/src/banner.js
```diff
@@ -0,0 +1,25 @@
...
+    tsserverPath = require.resolve(TSSERVER, {
+      paths: parseStringArray('--typeScriptProbeLocations'),
+    });
+  }
+  catch {}
+  const resolvedModules = modules.map(m => {
+    if (m === 'typescript') {
+      throw new Error(`'typescript' should never be used. Use '${TSSERVER}' instead.`)
+    }
+    return require(m === TSSERVER && tsserverPath || m);
```
ayazhafiz: Can you pull this out in a separate/longer conditional to more clearly explain what is being loaded? It took me a minute to realize this is equivalent to `m === TSSERVER && tsserverPath ? tsserverPath : m`.

package.json
kyliau: Good call! This actually broke the vscode dev workflow because rollup was not run.

server/src/banner.js
kyliau: done.

PR #406, opened 2019-10-09
ci: Check diff in format.sh only in CI mode

kyliau:
PR closes https://github.com/angular/vscode-ng-language-service/pull/405


kyliau:
Build on Travis CI would fail if any of the scripts fail, so ok to close #405.

Reference: https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase

PR #415, opened 2019-10-11
feat: Allow users to set @angular/language-service from vscode

kyliau:
This PR allows users to change the version of typescript and
@angular/language-service used by the server from vscode.

Instead of using process.cwd() (which is a very bad idea), the extension
client should instead look for typescript and @angular/language-service
in the workspace folders.
If there's no workspace folder, fallback to the bundled version.

Screenshot from 2019-10-10 17-06-23

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,17 @@
         "title": "Restart Angular Language server",
         "category": "Angular"
       }
-    ]
+    ],
+    "configuration": {
+      "title": "Angular Language Service",
+      "properties": {
+        "angular.ngdk": {
+          "type": ["string", "null"],
+          "default": null,
+          "description": "Specifies the folder path to @angular/language-service."
+        }
+      }
+    }
   },
   "activationEvents": [
     "onLanguage:html",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -23,7 +23,17 @@
         "title": "Restart Angular Language server",
         "category": "Angular"
       }
-    ]
+    ],
+    "configuration": {
+      "title": "Angular Language Service",
+      "properties": {
+        "angular.ngdk": {
```
ayazhafiz: What is `ngdk`? Also, can you add to the description that this points to the Language Service in node modules or falls back on the bundled version by default?

kyliau: It's not a very intuitive name, I must agree, but I also think it's good to align with TypeScript.
TypeScript has a config named `tsdk`, which I assume is short for "TypeScript development kit"
https://github.com/microsoft/vscode/blob/f3d4a67c2dddf45cb1636f959ae47ce47ead60d2/extensions/typescript-language-features/package.json#L119-L127

The description here will generate a tooltip that explains the field. I don't think we should mention fallback to the bundled version, because the very point of specifying `ngdk` is to override that behavior.

ayazhafiz: Is there anyway to mark it as optional, then? I just worry that some users may think this is a required configuration.

kyliau: With the default value set, it's already optional. Same with `typescript.tsdk`, by default it's null.

Issue #416, opened 2019-10-11
Intellisense for Angular Templates not working in v0.900.0-next.1


prasworld:
After the latest upgrade the angular intellisense has ceased to work. 
Cli version @angular/cli                       7.3.0
Platform Windows 
VS Code Version 1.39.1

iamcco:
Seems need `@angular/language-service >= v9.0.0` to work

Urastor:
Having the same problem. 

CLI version ~8.2.5
Platform Windows
VS Code Version 1.39.1
@angular/language-service 9.0.0-next.10

App is working fine using child router-outlet, but template gives error. 

![Anmerkung 2019-10-11 111055](https://user-images.githubusercontent.com/7266207/66639779-d0067d80-ec17-11e9-9139-c22f80690cf3.png)

kyliau:
Thanks for trying out the beta version! 
Could you please open the "Output" console in vscode and paste the logs here?
It should look something like this:

<img width="1792" alt="Screen Shot 2019-10-11 at 10 43 23 AM" src="https://user-images.githubusercontent.com/2941178/66672829-11952980-ec14-11e9-9c7f-5f1a45f5c933.png">

iamcco:
I have this log:

```
[Info  - 2:33:10 PM] Using typescript v3.5.3 from /Users/aioi/.vscode/extensions/angular.ng-template-0.900.0-next.1/node_modules/typescript
[Info  - 2:33:10 PM] Using @angular/language-service v9.0.0-next.10 from /Users/aioi/.vscode/extensions/angular.ng-template-0.900.0-next.1/server/node_modules/@angular/language-service
[Info  - 2:33:10 PM] Log file: /Users/aioi/Library/Application Support/Code/logs/20191015T142646/exthost1/Angular.ng-template/nglangsvc.log
[Info  - 2:33:18 PM] Disabling language service for /Users/aioi/workspace/pxn/anduin/tsconfig.json because it is not an Angular project.
```

It's weird the log `it is not an Angular project.`

iamcco:
![image](https://user-images.githubusercontent.com/5492542/66883357-3ddbec00-f000-11e9-9610-2abea96e9153.png)

I have `@angular/core` installed and it works with previous version。

kyliau:
@iamcco Do you mind sharing your `tsconfig.json` please? Is `node_modules` excluded?

iamcco:
The tsconfig.json

```jsonc
{
[...]

PR #423, opened 2019-10-18
fix: Check language service is not already disabled

kyliau:
When making the check to determine if a project is Angular, it could be
the case that the language service is already disabled. This happens
when there are files in the project that exceeded limit.

This PR fixes the issue by first checking if the language service is
already disabled. Also added much more information to the logs to help
users diagnose the problem. Messages are logged to both remote console
(vscode) and file so that the former could inform users in their editor,
and the latter makes it much easier to correlate the sequence of events
in both TypeScript and Angular.

Related: https://github.com/angular/vscode-ng-language-service/issues/416

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -83,13 +83,9 @@ export class Session {
         break;
       case ts.server.ProjectLoadingFinishEvent: {
         const {project} = event.data;
-        if (!isAngularProject(project)) {
-          project.disableLanguageService();
-          this.connection.console.info(`Disabling language service for ${
-              project.projectName} because it is not an Angular project.`);
-        }
-        this.connection.sendNotification(
-            projectLoadingNotification.finish, event.data.project.projectName);
+        // Disable language service if project is not Angular
+        this.checkIsAngularProject(project);
+        this.connection.sendNotification(projectLoadingNotification.finish, project.projectName);
         break;
       }
       case ts.server.ProjectsUpdatedInBackgroundEvent:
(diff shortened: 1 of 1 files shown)


server/src/session.ts
```diff
@@ -378,19 +374,53 @@ export class Session {
...
+   * Determine if the specified `project` is Angular, and disable the language
+   * service if not.
+   * @param project
+   */
+  private checkIsAngularProject(project: ts.server.Project) {
+    const NG_CORE = '@angular/core/core.d.ts';
+    const {projectName} = project;
+    if (!project.languageServiceEnabled) {
+      const msg = `Language service is already disabled for ${projectName}. ` +
+          `This could be due to non-TS files that exceeded limit. ` +
```
ayazhafiz: Can you specify what limit is exceeded here?

kyliau: done

server/src/session.ts
```diff
@@ -378,19 +374,54 @@ export class Session {
...
+   * Determine if the specified `project` is Angular, and disable the language
+   * service if not.
+   * @param project
+   */
+  private checkIsAngularProject(project: ts.server.Project) {
+    const NG_CORE = '@angular/core/core.d.ts';
+    const {projectName} = project;
+    if (!project.languageServiceEnabled) {
+      const msg = `Language service is already disabled for ${projectName}. ` +
+          `This could be due to non-TS files that exceeded limit (${
```
ayazhafiz:
```suggestion
          `This could be due to non-TS files that exceeded the size limit (${ts.server.maxProgramSizeForNonTsFiles} bytes).`
```

I think we should be as clear as possible; a large number with no annotation might be ambiguous. LGTM otherwise.

PR #425, opened 2019-10-18
feat: Allow users to set log verbosity in editor

kyliau:
This commit adds a vscode config `angular.log` that mirrors
`typescript.tsserver.log` for setting log verbosity. This allows users
to turn on verbose logging for debugging / diagnostic purpose.

It would also help us gather much more info when users report errors.

The default value is set to `terse`, which is minimal logging.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -14,26 +14,13 @@ import {registerCommands} from './commands';
 import {projectLoadingNotification} from './protocol';
 
 export function activate(context: vscode.ExtensionContext) {
-  // Log file does not yet exist on disk. It is up to the server to create the
-  // file.
-  const logFile = path.join(context.logPath, 'nglangsvc.log');
-  const ngProbeLocations = getProbeLocations('angular.ngdk', context.asAbsolutePath('server'));
-  const tsProbeLocations = getProbeLocations('typescript.tsdk', context.extensionPath);
   // If the extension is launched in debug mode then the debug server options are used
   // Otherwise the run options are used
   const serverOptions: lsp.ServerOptions = {
     run: {
       module: context.asAbsolutePath(path.join('server')),
       transport: lsp.TransportKind.ipc,
-      args: [
-        '--logFile',
-        logFile,
-        // TODO: Might want to turn off logging completely.
-        '--ngProbeLocations',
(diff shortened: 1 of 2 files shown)


client/src/extension.ts
```diff
@@ -143,3 +126,31 @@ function getProbeLocations(configName: string, bundled: string): string[] {
...
+/**
+ * Construct the arguments that's used to spawn the server process.
+ * @param ctx vscode extension context
+ * @param debug true if debug mode is on
+ */
+function constructArgs(ctx: vscode.ExtensionContext, debug: boolean): string[] {
+  const config = vscode.workspace.getConfiguration();
+  const args: string[] = [];
+
+  const ngLog = config.get('angular.log') as string;
```
ayazhafiz: Is the return value of `get` typed? If not, prefer pre-assignment typing:
```suggestion
  const ngLog: string = config.get('angular.log');
```

client/src/extension.ts
```diff
@@ -143,3 +126,31 @@ function getProbeLocations(configName: string, bundled: string): string[] {
...
+  if (ngLog !== 'off') {
+    // Log file does not yet exist on disk. It is up to the server to create the file.
+    const logFile = path.join(ctx.logPath, 'nglangsvc.log');
+    args.push('--logFile', logFile);
+    args.push('--logVerbosity', debug ? 'verbose' : ngLog);
+  }
+
+  const ngdk = config.get('angular.ngdk') as string | null;
```
ayazhafiz: same as above

client/src/extension.ts
```diff
@@ -143,3 +126,31 @@ function getProbeLocations(configName: string, bundled: string): string[] {
...
+  const ngProbeLocations = getProbeLocations(ngdk, ctx.asAbsolutePath('server'));
+  args.push('--ngProbeLocations', ngProbeLocations.join(','));
+
+  const tsdk = config.get('typescript.tsdk') as string | null;
```
ayazhafiz: same as above

client/src/extension.ts
kyliau: Good call! Your suggestion is what ts-style guide recommends. Thanks!

PR #431, opened 2019-10-21
fix: change minimal typescript version to 3.6.0.

andrius-pra:
This change will prevent this error in html file: `Failed to find project for /some/path/to/template.html`.

### DIFF
--- a/server/src/version_provider.ts
+++ b/server/src/version_provider.ts
@@ -31,40 +31,43 @@ function resolve(packageName: string, paths: string[]): NodeModule|undefined {
   }
 }
 
-function minVersion(nodeModule: NodeModule, minMajor: number): boolean {
+export function minVersion(nodeModule: NodeModule, minMajor: number, minMinor: number): boolean {
   if (!nodeModule.version) {
     return false;
   }
-  const [majorStr] = nodeModule.version.split('.');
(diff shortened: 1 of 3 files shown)


server/src/version_provider.ts
```diff
@@ -31,40 +31,43 @@ function resolve(packageName: string, paths: string[]): NodeModule|undefined {
...
       continue;
     }
-    if (minVersion(nodeModule, minMajor)) {
+    if (minVersion(nodeModule, minMajor, minMinor)) {
       return nodeModule;
     }
   }
   throw new Error(
-      `Failed to resolve '${packageName}' with minimum major version '${minMajor}' from ` +
+      `Failed to resolve '${packageName}' with minimum major '${minMajor}' and minor '${
```
ayazhafiz: I think we could word this error better. Maybe
```suggestion
      `Failed to resolve '${packageName}' with version >= ${minMajor}.${minMinor}.0
```

kyliau:
We'll have to rethink for a bit how to approach this.
The actual TS module is loaded in the shim `banner.js`.
The `ts` module resolved here is used to report the path, which could be different from the module actually loaded.
Perhaps the code should be refactored similar to ts plugin architecture, whereby the `ts` and `ng` module are resolved separately from the rest of the code and passed in as parameters.

Issue #432, opened 2019-10-28
Plugin process maxing out for ages on each file load


jpike88:
![Screen Shot 2019-10-28 at 12 12 08 pm](https://user-images.githubusercontent.com/9585787/67653528-57742080-f97c-11e9-8852-201facfc5b50.png)

^ This is happening practically non-stop, it's choking on files that don't even have angular templates enabled.

I'm uninstalling this plugin and just doing without, as it's rendering things unusable for me. I'm on a recent model Mac mini with 16gb of ram, plenty of resources and it's bringing it to its knees.

ayazhafiz:
What version of the plugin are you on?

kyliau:
@jpike88 Could you please provide the log files to help us debug this issues?
The location of the log file is shown when you open the output tab in vscode.
It'd look similar to this:
![Screenshot from 2019-09-18 14-52-29](https://user-images.githubusercontent.com/2941178/67700808-e88ecb80-f96b-11e9-8c62-5ba30caa2541.png)

jpike88:
I think I know the problem.

```107	918	44223	             /Applications/Visual Studio Code.app/Contents/Frameworks/Code Helper.app/Contents/MacOS/Code Helper /Users/joshpike/.vscode/extensions/angular.ng-template-0.900.0-next.6/server --logFile /Users/joshpike/Library/Application Support/Code/logs/20191029T115049/exthost1/Angular.ng-template/nglangsvc.log --logVerbosity terse --ngProbeLocations /Users/joshpike/Projects/aquipa,/Users/joshpike/.vscode/extensions/angular.ng-template-0.900.0-next.6/server --tsProbeLocations node_modules/typescript/lib,/Users/joshpike/Projects/aquipa,/Users/joshpike/.vscode/extensions/angular.ng-template-0.900.0-next.6 --node-ipc --clientProcessId=44214```

This process spikes on a single keypress. If I do 20 or 30 characters, the process just maxes out for ages, like 20 seconds.

Do you debounce any aspect of this plugin?

kyliau:
@jpike88 thanks for the logs! The logs did not give much info about the situation you encountered, but I think I should look into implementing [request cancellation](https://github.com/Microsoft/TypeScript/wiki/Standalone-Server-%28tsserver%29#cancellation).
This new extension started out as a prototype, so that part was skipped. It currently uses a null cancellation token.

[...]

PR #434, opened 2019-10-30
fix: Produce diagnostics only after user stops typing

kyliau:
This commit creates a timer to keep track of the last change made by the
user, and sends the diagnostics only after user stops typing.
Currently, it triggers for every change, and slows down the entire
operation.
A default delay of 200ms is consistent with TypeScript.

PR closes https://github.com/angular/vscode-ng-language-service/issues/432

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -38,6 +38,7 @@ const EMPTY_RANGE = lsp.Range.create(0, 0, 0, 0);
 export class Session {
   private readonly connection: lsp.IConnection;
   private readonly projectService: ProjectService;
+  private diagnosticsTimeout: NodeJS.Timeout|null = null;
 
   constructor(options: SessionOptions) {
     // Create a connection for the server. The connection uses Node's IPC as a transport.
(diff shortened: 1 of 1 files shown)


server/src/session.ts
```diff
@@ -91,16 +92,38 @@ export class Session {
...
+    // TypeScript. See
+    // https://github.com/microsoft/vscode/blob/7b944a16f52843b44cede123dd43ae36c0405dfd/extensions/typescript-language-features/src/features/bufferSyncSupport.ts#L493)
+    // A request is dropped if there's already a pending one.
+    if (this.hasPendingDiagnosticsRequest) {
+      return;
+    }
+    this.hasPendingDiagnosticsRequest = true;
+    setTimeout(() => {
+      this.hasPendingDiagnosticsRequest = false;
+      this.sendPendingDiagnostics(openFiles);
```
ayazhafiz: Regardless of what we do, I think the diagnostic should be sent immediately and only `hasPendingDiagnosticsRequest` is updated after 200ms. Otherwise we are waiting for even a diagnostic to be triggered.

kyliau:
I agree that diagnostics should be sent only after user stops typing. 
@andrius-pra also provided similar feedback [here](https://github.com/angular/vscode-ng-language-service/issues/432#issuecomment-548133221).
I've revised the implementation, PTAL.

In TypeScript, it's a little more involved because `getErr` is a multi-step process. When the diagnostics request is sent to the server, the server responds with a notification for **each** open file. This is more efficient, but it also requires the infrastructure to support cancelling a request that's in-flight. 
I think the current solution suffices for now, if not we could look into request cancellation in the next release.

jpike88:
im ready to test this

ayazhafiz:
Maybe we need to do this completions, etc. as well.

kyliau:
I'll merge this for now, and continue to investigate.

andrius-pra:
> The fix doesn't work in big project. ex: angular/angular-cli-stress-test

sorry for comment. it's not related to diagnostic messages. It works fine.

kyliau:
I tested the extension on a somewhat-large project: https://github.com/gothinkster/angular-realworld-example-app
it seems to be ok. Will do more testing.

jpike88:
Thanks, I have quite a large project so thanks for the load testing. My project is pretty big so every little bit helps. I’ll test new version and let you know

kyliau:
@jpike88 thank you so much!! I appreciate your patience and you helping us to stress test the new extension 🙏
I've released the extension as [v0.900.7](https://github.com/angular/vscode-ng-language-service/releases/tag/v0.900.0-next.7)

jpike88:
yeah definitely a big improvement. thanks again

PR #439, opened 2019-11-13
fix: Retrieve diagnostics for changed document only

kyliau:
Currently, when a document is changed, we retrieve diagnostics for
**all** open files. This could be very slow when there are many open
files. Instead, retrieve diagnostics for that one file only.

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -220,7 +220,7 @@ export class Session {
     if (!project || !project.languageServiceEnabled) {
       return;
     }
-    project.refreshDiagnostics();
+    this.triggerDiagnostics([scriptInfo.fileName]);
   }
 
   private onDidSaveTextDocument(params: lsp.DidSaveTextDocumentParams) {


ayazhafiz:
can we add hooks for opened documents?

kyliau:
> I don't like a new behavior. I expect that error will disappear when the code is fixed:
> diagnostics

Ahhh I forgot to take into account this use case where the fix for an error is not in the file itself but some other files.
We shall stick to the current implementation then, thank you for bringing this up.
TypeScript sorts the open files by most recently used, then retrieve diagnostics in that order. We could implement similar behavior.
https://github.com/microsoft/TypeScript/blob/ef0cca7d12a320bc98cefebbf4d391de36003149/src/server/session.ts#L2057-L2105

andrius-pra:
I think you can call `this.triggerDiagnostics` for changes in external file and `project.refreshDiagnostics` for changes in other files:
```
    if (scriptInfo.scriptKind === ts.ScriptKind.External) {
      this.triggerDiagnostics([scriptInfo.fileName]);
    } else {
      project.refreshDiagnostics();
    }
```

ayazhafiz:
> I think you can call `this.triggerDiagnostics` for changes in external file and `project.refreshDiagnostics` for changes in other files:
> 
> ```
>     if (scriptInfo.scriptKind === ts.ScriptKind.External) {
>       this.triggerDiagnostics([scriptInfo.fileName]);
>     } else {
>       project.refreshDiagnostics();
>     }
> ```

This would reload all diagnostics even in closed files, no?

andrius-pra:
`project.refreshDiagnostics();` will trigger diagnostics only for open files.

andrius-pra:
you can problemMatchers for closed files. ex: https://github.com/Genuitec/vscode-angular-cli-task-provider/blob/ea3346129ce1548494e74c0f3df835e0e3369ebe/package.json#L70-L101

PR #452, opened 2019-11-25
community: Update issue templates

kyliau:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
```
ayazhafiz: extra line?

kyliau:
> Do you think it would be helpful to add a note routing issues core to the language service, rather than the extension, to the monorepo?

Not many users are aware of the difference between the extension and the language service.
I think the monorepo is too large for us to track language service-specific issues effectively (I turned off notifications for @angular/angular because there's too much noise). So I think it's better to advertise this repository for issue tracking, but we will of course still keep tabs on the monorepo.

.github/ISSUE_TEMPLATE/bug_report.md
kyliau: fixed.

kyliau:
I also added a screenshot example.

PR #461, opened 2019-11-27
feat: add syntax highlighting for embedded HTML in TypeScript

ayazhafiz:
Injects an HTML TextMate grammar in Angular component `template`
decorator properties, in TypeScript files.

The grammar is enabled in decorators with a `template` property whose
key value is a string surrounded by any number of parantheses or
whitespace.

This project owes a great deal to
https://github.com/natewallace/angular2-inline, who first enabled
Angular inline syntax highlighting.

syntax highlighting

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,19 @@
           "description": "Enables logging of the Angular server to a file. This log can be used to diagnose Angular Server issues. The log may contain file paths, source code, and other potentially sensitive information from your project."
         }
       }
-    }
+    },
+    "grammars": [
+      {
+        "path": "./syntaxes/template.ng.json",
+        "scopeName": "template.ng",
+        "injectTo": [
+          "source.ts"
+        ],
+        "embeddedLanguages": {
+          "source.html": "html"
+        }
+      }
+    ]
   },
   "activationEvents": [
     "onLanguage:html",
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -16,6 +16,8 @@ cp package.json yarn.lock angular.png CHANGELOG.md README.md dist
 cp client/package.json client/yarn.lock dist/client
 # Copy files to server directory
 cp server/package.json server/yarn.lock server/README.md dist/server
+# Copy files to syntaxes directory
+cp -R syntaxes dist/syntaxes
 
 pushd dist
 yarn install --production --ignore-scripts
--- a/syntaxes/template.ng.json
+++ b/syntaxes/template.ng.json
@@ -1,0 +1,73 @@
+{
+  "scopeName": "template.ng",
+  "injectionSelector": "L:meta.decorator.ts",
+  "patterns": [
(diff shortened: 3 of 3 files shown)


kyliau:
Nice work!! I was wondering if we covered all use cases, and you covered template literals as well as plain string. Thank you =D

andrius-pra:
The syntax highlighting disappears when I have template property in decorator:
bug

kyliau:
> The syntax highlighting disappears when I have template property in decorator:
> bug

To be more specific, it's syntax highlighting for any TypeScript code **below** the `template` that disappears.

The scope for any code below `template` becomes
- `meta.objectliteral.ts`
- `meta.decorator.ts`
- `source.ts`

ayazhafiz:
Yep, that's my bad - I forgot to specify a closing match for the `template` property. Can you retry the highlighter?

kyliau:
This is so awesome! I've attached a screenshot so that it's easier for us to showcase this function.

syntax highlighting

ayazhafiz:
I just found https://github.com/PanAeon/vscode-tmgrammar-test, looking into writing tests with it now.

ayazhafiz:
> I just found https://github.com/PanAeon/vscode-tmgrammar-test, looking into writing tests with it now.

Okay, after some work, looks like this isn't a viable option right now because `tmgrammar-test` cannot match against more than one grammar currently, causing grammars outside the scope of the ones provided in this repo (for instance, VSCode TS and HTML grammars) to not be analyzable (https://github.com/PanAeon/vscode-tmgrammar-test/issues/3). In particular, we can only build partially-useful tests - some scopes can be matched, but go far enough in on a scope that relies on an external grammar, and the tests stop working.

kyliau:
> > I just found https://github.com/PanAeon/vscode-tmgrammar-test, looking into writing tests with it now.
> 
> Okay, after some work, looks like this isn't a viable option right now because `tmgrammar-test` cannot match against more than one grammar currently, causing grammars outside the scope of the ones provided in this repo (for instance, VSCode TS and HTML grammars) to not be analyzable ([PanAeon/vscode-tmgrammar-test#3](https://github.com/PanAeon/vscode-tmgrammar-test/issues/3)). In particular, we can only build partially-useful tests - some scopes can be matched, but go far enough in on a scope that relies on an external grammar, and the tests stop working.

Since we don't anticipate the grammar to change frequently, I think some minimal test is good for now.

andrius-pra:
The `build.sh` does not include `syntaxes/template.ng.json` into `ngls.vsix`. Can you also update build.sh to include `syntaxes/template.ng.json` into .vsix package

ayazhafiz:
> The `build.sh` does not include `syntaxes/template.ng.json` into `ngls.vsix`. Can you also update build.sh to include `syntaxes/template.ng.json` into .vsix package

Done

PR #467, opened 2019-11-29
test: add tests for inline template textmate grammar

ayazhafiz:
Adds tests for the TypeScript inline Angular template grammar via the
[vscode-tmgrammar-test](https://github.com/PanAeon/vscode-tmgrammar-test) framework, which [now supports loading multiple grammars](https://github.com/PanAeon/vscode-tmgrammar-test/commit/e2c4e86cd0f365a106125f3df03208d90b11a72e)
as needed by this project.

A dummy HTML grammar is used for inline templates, as currently we are
only concerned with recognization of the inline template.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ script:
   - ./scripts/build.sh
   - ./scripts/test.sh
   - ./scripts/format.sh
+  - ./scripts/syntax.sh
 
 after_script:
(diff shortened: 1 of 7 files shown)


ayazhafiz:
Unfortunately, we can't format it because the test runner requires comments to start at the first index of a line. Instead, I've excluded the test file from clang-format. Let me know if that works!

PR #472, opened 2019-12-02
fix(syntaxes): do not syntax highlight commented template properties

ayazhafiz:
Prevents syntax highlighting inline templates by excluding the grammar
injection on decorators with comments.

The grammar test framework does not currently support specifying grammar
rule exclusions (https://github.com/PanAeon/vscode-tmgrammar-test/issues/6),
so instead the test cases are moved to snapshot tests
that demonstrate that no other grammars are added on excluded cases.

### DIFF
--- a/scripts/syntax.sh
+++ b/scripts/syntax.sh
@@ -15,8 +15,5 @@ ARGS=$(cat<<ARGS
 ARGS
 )
 
-# Unit tests
-yarn vscode-tmgrammar-test $ARGS
-
 # Snapshot tests
 yarn vscode-tmgrammar-snap $ARGS "$@"
--- a/scripts/syntax.sh
+++ b/scripts/syntax.sh
@@ -2,7 +2,7 @@
 # Usage:
 #   scripts/syntax.sh [-u]
 #
-# Parameters
+# Arguments:
 #   -u update snapshot files
 
 set -ex -o pipefail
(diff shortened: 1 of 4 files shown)


scripts/syntax.sh
```diff
@@ -3,9 +3,15 @@
...
-# Template syntax tests
-yarn vscode-tmgrammar-test \
-  -s template.ng \
-  -g syntaxes/template.ng.json $DUMMY_GRAMMARS \
-  -t "syntaxes/test/**/*.ts"
+# Unit tests
+yarn vscode-tmgrammar-test $ARGS
+
+# Snapshot tests
+yarn vscode-tmgrammar-snap $ARGS
```
kyliau: Does this generate the snapshot or compares the output with an existing snapshot?
Could you please add a comment somewhere on how to update the snapshot in case a test fails?

ayazhafiz: It compares the output with an existing snapshot. I've updated the script to take a '-u' argument to update the snapshots.

scripts/syntax.sh
```diff
@@ -1,11 +1,22 @@
...
+ARGS
+)
 
-# Template syntax tests
-yarn vscode-tmgrammar-test \
-  -s template.ng \
-  -g syntaxes/template.ng.json $DUMMY_GRAMMARS \
-  -t "syntaxes/test/**/*.ts"
+# Unit tests
+yarn vscode-tmgrammar-test $ARGS
```
kyliau: I'm a little confused, now that `inline_template.ts` no longer has those inline annotations, how's the unit test different from snapshot test?

ayazhafiz: It's not. I just left the lines in there in case we add unit tests later on. Would you prefer removing it?

kyliau: ah, I see. Let's remove it until it's actually needed. Thank you!

ayazhafiz: 👍

Issue #469, opened 2019-12-01
The Language server is crashing when templateUrl points to non-existing file


andrius-pra:
**Describe the bug**

The Language server is restarting when templateUrl points to non-existing file.

**To Reproduce**
1. Update to version 0.900.0
2. Open any angular project
4. change template url to non-existing filename.

**Expected behavior**

Expected to get `URL does not point to a valid file` diagnostic message.

**Logs**

1. Console output:
```
[Info  - 22:31:47] Connection to server got closed. Server will restart.
[Info  - 22:31:47] Angular language server process ID: 18316
[Info  - 22:31:47] Using typescript v3.6.4 from c:\Users\Andrius\.vscode\extensions\angular.ng-template-0.900.0\node_modules\typescript\lib\tsserverlibrary.js
[Info  - 22:31:47] Using @angular/language-service v9.0.0-rc.3 from c:\Users\Andrius\.vscode\extensions\angular.ng-template-0.900.0\server\node_modules\@angular\language-service\bundles\language-service.umd.js
[Info  - 22:31:47] Log file: c:\Users\Andrius\AppData\Roaming\Code\logs\20191202T003053\exthost1\Angular.ng-template\nglangsvc.log
[Error - 22:31:52] TypeError: Cannot read property 'attachToProject' of undefined
    at c:\Users\Andrius\.vscode\extensions\angular.ng-template-0.900.0\node_modules\typescript\lib\tsserverlibrary.js:129871:32
    at Object.enumerateInsertsAndDeletes (c:\Users\Andrius\.vscode\extensions\angular.ng-template-0.900.0\node_modules\typescript\lib\tsserverlibrary.js:2003:13)
    at ConfiguredProject.Project.updateGraphWorker (c:\Users\Andrius\.vscode\extensions\angular.ng-template-0.900.0\node_modules\typescript\lib\tsserverlibrary.js:129865:20)
    at ConfiguredProject.Project.updateGraph (c:\Users\Andrius\.vscode\extensions\angular.ng-template-0.900.0\node_modules\typescript\lib\tsserverlibrary.js:129745:42)
    at ConfiguredProject.updateGraph (c:\Users\Andrius\.vscode\extensions\angular.ng-template-0.900.0\node_modules\typescript\lib\tsserverlibrary.js:130316:63)
[...]

Issue #460, opened 2019-11-27
index value is not number error in template


sheikalthaf:
**Describe the bug**

Showing error for index value from ngFor is not numeric value.
```
(variable) i
Expected a numeric typeng(0)
```
**To Reproduce**

**Expected behavior**

**Logs**

Please attach two logs:

1. Console output (Go to toolbar --> View --> Output, and copy the contents)
2. Log file (Path is printed to the console output at startup)

**Screenshots**

![image](https://user-images.githubusercontent.com/22526247/69695115-29ac0400-1101-11ea-9026-33199da6c072.png)

**Additional context**

ChamaraWP:
I getting same kind of error for the *ngIf check. I didn't get this before
![Capturess](https://user-images.githubusercontent.com/21265782/69695793-93c5a880-1103-11ea-90e3-c3c2c2fb9946.PNG)

migel-wimtore:
I am getting the same error when interpolating the index variable from an `ngFor` and trying to add a number to it. For example:

```
<ng-container *ngFor="let item of items; index as i">

{{i + 1}}

</ng-container>
```

I have noticed that if the keyvalue pipe is used in the ngFor declaration, the error does not appear. For example:

```
<ng-container *ngFor="let item of items | keyvalue; index as i">

{{i + 1}}

</ng-container>
```

kombasltd:
Same error here and somehow this ist new after a VS Code update and maybe an update of "Angular Files". I was wondering because this was old code which now throws this error, but serves and builds fine.

```
<div *ngFor="let product of productItems; let i = index">
      <div *ngIf="i % 3 === 0 " class="card-deck mt-2">
        <div class="card" *ngFor="let item of productItems.slice(i,i+3)">
...
```

The variable "i" in lines 2 and 3 are marked as an error "Expected a numeric type ng(0)".

p4309027:
in my case it is saying: 

`Expected a number type ng(0)`

for following code
```
<div *ngFor="let file of files; let i = index;">
   <p>File #{{i+1}}: {{file.name}}</p>
</div>
....
```

masterlavidobrasil:
No meu caso usei conversão via pipe {{ (index | number) + 1 }} e parou o problema

wesleyxavier:
No meu caso usei conversão via pipe {{ (index | number) + 1 }} e parou o problema

PR #479, opened 2019-12-06
docs: Update changelog for 0.900.1

kyliau:


kyliau:
I added #460 but #469 is not yet in rc.5 based on the [CHANGELOG](https://github.com/angular/angular/blob/master/CHANGELOG.md) in angular repo

We ought to figure out a better to automatically update the change log ..

ayazhafiz:
Maybe an action after a PR is closed? Or a label we can add to issues/PRs?

Do you know how the Angular monorepo does it?

PR #504, opened 2019-12-16
refactor: fly-by cleanup; merge terse and default logger branches

ayazhafiz:


server/src/logger.ts
```diff
@@ -53,6 +50,7 @@ export function createLogger(options: LoggerOptions): Logger {
     case 'normal':
       logLevel = ts.server.LogLevel.normal;
       break;
+    case 'terse':
```
kyliau: oops, sorry, I meant to provide suggestions, but the commit was added to your branch.
I meant to suggest adding `terse` to the default case so that it is more obvious all the flags that the program supports.

PR #506, opened 2019-12-17
refactor: rename template syntax to inline_template

ayazhafiz:
This commit performs a minor refactor, changing the TextMate inline
template syntax file name and scopes to explicitly be `inline_template`.
This is in preparation for adding a TextMate grammar for CSS styles in
`@Component` `styles` properties.


scripts/syntax.sh
```diff
@@ -9,7 +9,7 @@ set -ex -o pipefail
 
 DUMMY_GRAMMARS=$(find syntaxes/test -name '*-dummy.json' -exec echo "-g {}" \; | tr '\n' ' ')
 ARGS=$(cat<<ARGS
-  -s template.ng 
+  -s inline-template.ng 
   -g syntaxes/template.ng.json $DUMMY_GRAMMARS
```
kyliau: gotta update this filename as well

PR #510, opened 2019-12-17
refactor: modularize syntax grammar tests

ayazhafiz:
As we add more TextMate grammars and tests, the bash script we currently
have would not suffice because for an singular snapshot or unit tests,
the TextMate grammar testing software [we
use](https://github.com/PanAeon/vscode-tmgrammar-test) requires

1. The scope of the grammar to test
2. The grammar specification files
3. The test file itself

To conviniently express these requirements for each test, this commit
adds a test runner written in JavaScript with an array specifying the
layout of each test.


ayazhafiz:
Should we pull out the array of tests to a JSON file?

scripts/syntax.js
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env node
```
kyliau: Instead of writing this as a script, consider structuring the tests using Jasmine. 
The LSP end-to-end test is a good example of this.
This would allow us to achieve two things:
1. Use standardized testing procedures
2. Use TypeScript throughout the codebase

What do you think?

ayazhafiz: Sounds good; I'll move it.

kyliau:
> Should we pull out the array of tests to a JSON file?
> 
> Update: moved to a JSON file. Turns out TypeScript can typecheck JSON files!

Wow, I didn't know that! It's even better!

Issue #483, opened 2019-12-07
Syntax Highlighting for Angular Template Syntax


ayazhafiz:
**Is your feature request related to a problem? Please describe.**

As an extension of the new Textmate grammar provided by this extension, we should provide a grammar for the [Angular template syntax](https://angular.io/guide/template-syntax).

**Additional context**

See https://github.com/angular/vscode-ng-language-service/issues/395 for the inline template syntax highlighting discussion.

Running list of features:

- [x] Inline template recognition (#461)
- [x] Embedded JS/CSS in a template (#513)
- [x] Interpolations (#516)
- [x] Bound properties (#524)
- [x] Bound events (#543)
- [x] Two way bindings (#566)
- [x] Structural directives/Microsyntax (#567)

jpike88:
@kyliau some useful CSS examples to cover:

![68101848-52761a80-ff02-11e9-8c07-681380946ce0](https://user-images.githubusercontent.com/9585787/70963097-16fe6c80-20ba-11ea-819b-7faf0fb78a9a.png)
![68099924-d4ad1180-fef7-11e9-8b4c-0a36e983da0b](https://user-images.githubusercontent.com/9585787/70963099-18c83000-20ba-11ea-950d-0f80dbdc6a27.png)
![Screen Shot 2019-12-17 at 10 43 49 am](https://user-images.githubusercontent.com/9585787/70963114-27aee280-20ba-11ea-9a4f-221083441a0e.png)

ayazhafiz:
@jpike88 I can't seem to reproduce your CSS examples in a template or CSS file. Are they located elsewhere?

jpike88:
These are inline templates, in a @component definition

ayazhafiz:
@jpike88 do you have the [angular2-inline](https://github.com/natewallace/angular2-inline) extension installed? I see what you're getting with the CSS styles, but only when that extension is enabled. If you turn off that extension but keep the language service extension enabled, you should see correct CSS highlighting in an inline template (styles in a `styles` property on a Component won't be, though).

jpike88:
@ayazhafiz you're right, the inline extension did modify colours. but its worse in some ways with just the angular extension. new screenshots:

![Screen Shot 2019-12-18 at 12 15 43 pm](https://user-images.githubusercontent.com/9585787/71057683-2354f900-2190-11ea-8351-1069c548153f.png)
[...]

PR #512, opened 2019-12-18
feat: add textmate grammar for inline CSS styles

ayazhafiz:
Adds a TextMate grammar for inline styles in a component.

Unfortunately, we cannot merge this grammar conviniently with the one
specified for inline templates, because though the directions of grammar
recurse are similar for both, they apply different grammars for
matches. For instance, an inline template and style both match a string
literal, but the former applies an HTML grammar while the later applies
a CSS grammar. To distinguish between the two, we have to create two
different grammar recurse paths at the closest divergence of the
grammars (for convinience, at `template` and `style` properties in a
`@Component` definition).

Attached to the PR for this commit are images demonstrating this change.
The second image shows that commenting out the style property applies
the regular comment grammar to the entire style string. For some reason,
however, running a commented style through the snapshot tests leads to
an incorrect and undeterministic result. This is likely an issue with the
test framework; I will investigate and update the tests in a future PR.

Partially addresses #483 ([comment](https://github.com/angular/vscode-ng-language-service/issues/483#issuecomment-566367049))

Screen Shot 2019-12-17 at 6 37 35 PM
Screen Shot 2019-12-17 at 7 14 53 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,10 +54,14 @@
         "scopeName": "inline-template.ng",
         "injectTo": [
           "source.ts"
-        ],
-        "embeddedLanguages": {
-          "source.html": "html"
-        }
+        ]
+      },
+      {
+        "path": "./syntaxes/inline-styles.json",
+        "scopeName": "inline-styles.ng",
+        "injectTo": [
+          "source.ts"
+        ]
       }
     ]
   },
--- a/package.json
+++ b/package.json
@@ -54,14 +54,20 @@
         "scopeName": "inline-template.ng",
         "injectTo": [
           "source.ts"
-        ]
+        ],
+        "embeddedLanguages": {
+          "source.html": "html"
+        }
       },
       {
         "path": "./syntaxes/inline-styles.json",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -54,10 +54,14 @@
         "scopeName": "inline-template.ng",
         "injectTo": [
           "source.ts"
-        ],
-        "embeddedLanguages": {
-          "source.html": "html"
```
kyliau: is this no longer needed?

ayazhafiz: Oh yeah. I removed this because there was a bug to fix with the embedded languages, but I'll do it in another PR. added back

kyliau:
looks like the golden files need to be updated.
Is there a doc somewhere that explains how to generate the goldens?

ayazhafiz:
Fixed. The documentation for updating is in the driver source file:

https://github.com/angular/vscode-ng-language-service/blob/d6c0ff18bb0d7c565d4b16cf30d88bfa2fa6f65c/syntaxes/test/driver.ts#L1-L6

PR #516, opened 2019-12-18
feat: add syntax highlighting grammar for interpolations

ayazhafiz:
Adds a grammar for Angular interpolations. Because the grammar relies
only on regex expressions, we cannot currently support any interpolation
delimiters other than the default `{{`/`}}` -- to do more would require
a more sophisticated highlighting solution.

For convinience, everything inside an interpolation is treated as
JavaScript syntax, even though the expressions available in an Angular
interpolations are a distinct subset of JavaScript.

Attached to this PR are several before/after images of this change.

Partially addresses #483.

| | Before   | After |
|:---:| :---:       |    :----:   | 
|Inline template| Screen Shot 2019-12-18 at 11 23 55 AM| Screen Shot 2019-12-18 at 11 24 13 AM |
|External template|Screen Shot 2019-12-18 at 11 23 27 AM | Screen Shot 2019-12-18 at 11 22 57 AM

### DIFF
--- a/syntaxes/test/data/inline-template.ts.snap
+++ b/syntaxes/test/data/inline-template.ts.snap
@@ -1,10 +1,12 @@
 >/* clang-format off */
 #^^^^^^^^^^^^^^^^^^^^^^^ inline-template.ng
 >
+>/* Inline template recognition tests */
+#^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ inline-template.ng
 >@Component({
 #^^^^^^^^^^^^^ inline-template.ng
->//// Property key/value test
-#^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ inline-template.ng
+>  //// Property key/value test
+#^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ inline-template.ng
 >  template: '<div></div>',
 #^^ inline-template.ng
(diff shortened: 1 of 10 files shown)


syntaxes/test/data/inline-template.ts.snap
```diff
@@ -93,4 +95,26 @@
 #^^^ inline-template.ng
 >export class TMComponent{}
 #^^^^^^^^^^^^^^^^^^^^^^^^^^^ inline-template.ng
+>
+>/* Template syntax tests */
+#^^^^^^^^^^^^^^^^^^^^^^^^^^^^ inline-template.ng
```
kyliau: Could you please explain why regular TypeScript nodes are also labelled `inline-template.ng`?

ayazhafiz: This seems to be a quirk of the testing framework inherited from TextMate. It require a base scope for the grammar language (in this case, `inline-template.ng`), which becomes the base grammar name of everything the grammar is applied on. More specific grammar names are stacked as the grammar proceeds.

kyliau: ah, I see. We might want to document that in a README.

ayazhafiz: I will add it to the documentation in a next PR.

PR #529, opened 2019-12-24
fix: whitelist all html elements

### LINKED ISSUE #526: Weird HTML syntax highlighting issue
Latest version of everything, see screenshot.

Screen Shot 2019-12-23 at 8 32 03 PM

ayazhafiz:
The `html.basic` grammar whitelists only W3 and custom HTML elements.
Some Angular users may use elements that do not have the custom format
(hyphen-separated), so just whitelist all elements. Fixes cases like
`<popover></popover>` being highlighted incorrectly.

Closes #526.

### DIFF
--- a/syntaxes/inline-template.json
+++ b/syntaxes/inline-template.json
@@ -67,7 +67,7 @@
       "contentName": "text.html",
       "patterns": [
         {
-          "include": "text.html.basic"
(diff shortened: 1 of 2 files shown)


jpike88:
@kyliau when are you back

PR #540, opened 2020-01-10
fix: strictly define characters in property binding

### LINKED ISSUE #539: let ... of parsing broken in HTML editor
The new version v0.900.4 does no longer parse `let ... of` expressions correctly - the HTML editor syntax coloring is broken for the rest of the file. Maybe ternary expressions in the `let ... of` expression cause this, see below.

Here is an (almost) minimal HTML source reproducing this issue:

```
<select class="form-control" [(ngModel)]="selectedApp" name="app" #app>
    <option *ngFor="let a of (someCondition ? myApps : [])" [ngValue]="a">{{ a.name }}</option>
</select>
```

And here is how this is rendered in the editor:

image

1. Upgrade Angular Language Service to v0.900.4
2. Create an HTML file with above contents

**Expected behavior**

The code should be rendered this way (output from v0.900.3):

image

**Additional context**

When downgrading the extension to v0.900.3, everything works fine.

I strongly assume that this "bugfix" introduced this behavior:

https://github.com/angular/angular/issues/34473

But I cannot identify if the associated angular change is faulty, or if the bug is within the vscode extension.

Please also note the wrong rendering of the `[(ngmodel)]` value. I am not sure if this is related or not.

ayazhafiz:
Avoids whitelisting all propery binding names, so that proximal but
seperate brackets aren't treated as overlapping of the same property
binding.

The screenshot attached to this PR demonstrates the fix - in previous
versions, `[1, 2, 3]" [ngValue]` would be treated as one property
binding name.

Fixes #539

Screen Shot 2020-01-10 at 7 50 04 AM

### DIFF
--- a/syntaxes/template.json
+++ b/syntaxes/template.json
@@ -32,7 +32,7 @@
     },
 
     "propertyBinding": {
-      "begin": "(\\[[ a-zA-Z0-9\\._\\-@]*\\])(=)(\"|')",
+      "begin": "(\\[[ a-zA-Z0-9\\._\\-@$]*\\])(=)(\"|')",
       "beginCaptures": {
         "1": {
           "name": "entity.other.attribute-name.html"
--- a/syntaxes/template.json
+++ b/syntaxes/template.json
@@ -32,7 +32,7 @@
     },
 
     "propertyBinding": {
-      "begin": "(\\[.*\\])(=)(\"|')",
+      "begin": "(\\[[ a-zA-Z0-9\\._\\-@]*\\])(=)(\"|')",
       "beginCaptures": {
         "1": {
           "name": "entity.other.attribute-name.html"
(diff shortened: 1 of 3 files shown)


syntaxes/template.json
```diff
@@ -32,7 +32,7 @@
     },
 
     "propertyBinding": {
-      "begin": "(\\[.*\\])(=)(\"|')",
+      "begin": "(\\[[ a-zA-Z0-9\\._\\-@]*\\])(=)(\"|')",
```
kyliau: This would not recognize the banana syntax, would it?
Could you please add a test case for that?

ayazhafiz: No, it won't. Do you mean to add a test case that it doesn't support the banana syntax?

dannymcgee: I think this is missing a `$` character. Not used super often (except for observables which you probably wouldn't want to property bind), but probably still worth including since it's easy to type and a valid character for identifiers.

ayazhafiz: True: https://github.com/angular/angular/blob/master/packages/language-service/src/completions.ts#L75

ayazhafiz:
I like the character rule Danny's PR more, closing in favor of that.

PR #542, opened 2020-01-12
feat: more detailed grammar scopes for template property binding syntax

### LINKED ISSUE #539: let ... of parsing broken in HTML editor

dannymcgee:
#### Changes:
- Strictly defined characters in the attribute name pattern (fixes #539)
- Reversed the order of `string.quoted.html` and `punctuation.definition.string.begin/end.html` scopes (TM grammar scopes go from less specific on the left to more specific on the right)

#### Additions:
- **Angular bindings** given a unique scope (`ng-binding.html`) — this scope covers the entire key/value pair, analogous to the HTML grammar's `meta.attribute.<attribute-name>.html` scope
- **Angular binding attribute names** given a unique scope in addition to `entity.other.attribute-name.html` (`entity.other.ng-binding-name.html`) — this scope covers the entire _key_, including punctuation
- **Punctuation** in Angular binding keys given unique scopes:
    - **Brackets and parens** get `punctuation.definition.ng-binding-name.begin/end.html` scopes
    - **Dots** in the attribute name get `punctuation.accessor.html` scope
- Updated template snapshots and added test cases to cover all the changes made above

### DIFF
--- a/syntaxes/template.json
+++ b/syntaxes/template.json
@@ -32,7 +32,7 @@
     },
 
     "ngBinding": {
-      "begin": "([\\[\\(]{1,2})(@?[-_a-zA-Z0-9.$]*)([\\]\\)]{1,2})(=)([\"'])",
+      "begin": "([\\[\\(]{1,2})(?:\\s*)(@?[-_a-zA-Z0-9.$]*)(?:\\s*)([\\]\\)]{1,2})(=)([\"'])",
       "beginCaptures": {
         "1": {
           "name": "punctuation.definition.ng-binding.begin.html"
--- a/syntaxes/template.json
+++ b/syntaxes/template.json
@@ -6,7 +6,7 @@
       "include": "#interpolation"
     },
     {
-      "include": "#ngBinding"
+      "include": "#propertyBinding"
     }
   ],
   "repository": {
@@ -31,35 +31,25 @@
       ]
     },
 
-    "ngBinding": {
-      "begin": "(([\\[\\(]{1,2})(?:\\s*)(@?[-_a-zA-Z0-9.$]*)(?:\\s*)([\\]\\)]{1,2}))(=)([\"'])",
+    "propertyBinding": {
+      "begin": "(\\[\\s*@?[-_a-zA-Z0-9.$]*\\s*])(=)([\"'])",
       "beginCaptures": {
         "1": {
-          "name": "entity.other.attribute-name.html entity.other.ng-binding-name.html"
-        },
-        "2": {
-          "name": "punctuation.definition.ng-binding-name.begin.html"
-        },
-        "3": {
           "name": "entity.other.attribute-name.html entity.other.ng-binding-name.html",
           "patterns": [
             {
-              "match": "\\.",
-              "name": "punctuation.accessor.html"
+              "include": "#bindingKey"
             }
           ]
         },
-        "4": {
-          "name": "punctuation.definition.ng-binding-name.end.html"
-        },
-        "5": {
+        "2": {
           "name": "punctuation.separator.key-value.html"
         },
-        "6": {
+        "3": {
           "name": "string.quoted.html punctuation.definition.string.begin.html"
         }
       },
-      "end": "\\6",
+      "end": "\\3",
       "endCaptures": {
         "0": {
           "name": "string.quoted.html punctuation.definition.string.end.html"
--- a/syntaxes/template.json
+++ b/syntaxes/template.json
@@ -35,7 +35,7 @@
       "begin": "(\\[\\s*@?[-_a-zA-Z0-9.$]*\\s*])(=)([\"'])",
       "beginCaptures": {
         "1": {
-          "name": "entity.other.attribute-name.html entity.other.ng-binding-name.html",
+          "name": "entity.other.attribute-name.html entity.other.ng-binding-name.property.html",
           "patterns": [
             {
               "include": "#bindingKey"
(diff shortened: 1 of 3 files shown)


dannymcgee:
@googlebot I signed it!

syntaxes/template.json
```diff
@@ -31,25 +31,41 @@
       ]
     },
 
-    "propertyBinding": {
-      "begin": "(\\[.*\\])(=)(\"|')",
+    "ngBinding": {
+      "begin": "(([\\[\\(]{1,2})(?:\\s*)(@?[-_a-zA-Z0-9.$]*)(?:\\s*)([\\]\\)]{1,2}))(=)([\"'])",
```
ayazhafiz: I would rather keep three separate rules for property, event, and two-way bindings, because a merged pattern will run into issues of matching mixed patterns, like `[prop)`.

syntaxes/template.json
```diff
@@ -31,25 +31,41 @@
...
       "beginCaptures": {
         "1": {
-          "name": "entity.other.attribute-name.html"
+          "name": "entity.other.attribute-name.html entity.other.ng-binding-name.html"
         },
         "2": {
-          "name": "punctuation.separator.key-value.html"
+          "name": "punctuation.definition.ng-binding-name.begin.html"
```
ayazhafiz: What is the advantage of this custom scope name?

syntaxes/template.json
dannymcgee: @ayazhafiz Do you think I should go ahead and split them up in this PR, or keep this PR to just property binding and add event/two-way binding in future PRs? I don't mind doing the work here if you wouldn't object to a bigger diff

ayazhafiz: Let's keep this one to property binding and add the rest in separate PRs ))

syntaxes/template.json
dannymcgee: @ayazhafiz It provides a hook for theme authors to give a unique token color to the punctuation, or to Angular bindings vs regular HTML attributes, etc. The `entity.other.attribute-name` scope is kept as a fallback so themes that haven't gone out of their way to color Angular bindings won't have missing colors.

For example, this is how my theme handles the test cases in this branch:

Angular template syntax highlighting

syntaxes/template.json
dannymcgee: @ayazhafiz This should be taken care of now

dannymcgee:
> Thanks for the PR, Danny. I think we should stick to individual rules for property/event/two-way bindings. Could you add some tests for validating/invalidating property binding names?

Added some additional test cases and removed the ones for event/two-way binding. Let me know if there's anything I missed!

dannymcgee:
> LGTM aside just the two comments.
@ayazhafiz Should all be taken care of! I left the one comment unresolved so you could review my explanation, but feel free to resolve it if you're satisfied

syntaxes/template.json
```diff
@@ -32,30 +32,60 @@
     },
 
     "propertyBinding": {
-      "begin": "(\\[.*\\])(=)(\"|')",
+      "begin": "(\\[\\s*@?[-_a-zA-Z0-9.$]*\\s*])(=)([\"'])",
       "beginCaptures": {
         "1": {
-          "name": "entity.other.attribute-name.html"
+          "name": "entity.other.attribute-name.html entity.other.ng-binding-name.html",
```
ayazhafiz: What do you think of making the scope name more specific, like `ng-property-binding-name`?

syntaxes/template.json
```diff
@@ -32,30 +32,60 @@
...
         }
       },
       "end": "\\3",
       "endCaptures": {
         "0": {
-          "name": "punctuation.definition.string.end.html string.quoted.html"
+          "name": "string.quoted.html punctuation.definition.string.end.html"
         }
       },
+      "name": "ng-binding.html",
```
ayazhafiz: Same as above

syntaxes/template.json
dannymcgee: Yeah, that's a good idea now that this scope isn't generic anymore.

dannymcgee: Updated! 0a254fe6fcd05a2f9fd60473328ed83ddd7d72ef

syntaxes/template.json
dannymcgee: Updated! 0a254fe6fcd05a2f9fd60473328ed83ddd7d72ef

For this I took a cue from the HTML grammar and also added a `meta` tag to the scope (explanation [here](https://macromates.com/manual/en/language_grammars#naming_conventions)). The HTML grammar also inserts the property name into their equivalent scope with a backreference, which is something we can think about doing for the built-in directives like `ngStyle`/`ngClass` later on down the road.

PR #553, opened 2020-01-15
deps: update vscode-languageclient & languageserver to 6.0.0

ayazhafiz:


kyliau:
@ayazhafiz could you please rebase? thank you

server/package.json
```diff
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "@angular/language-service": "9.0.0-rc.9",
+    "vscode-languageclient": "6.0.0",
```
kyliau: why do we need language client in the server?

ayazhafiz: We don’t. I just chose to update both at once.

kyliau: huh? but this is not updating the dep, it's *adding* a new dep. I'm wondering why this dep is needed.

kyliau: Did you mean to update `vscode-languageserver`, which is the dep below?

ayazhafiz: Apologies. Should be fixed now.

server/src/session.ts
```diff
@@ -229,9 +229,12 @@ export class Session {
       return;
     }
     for (const change of contentChanges) {
-      if (change.range) {
+      if ('range' in change) {
         const [start, end] = lspRangeToTsPositions(scriptInfo, change.range);
         scriptInfo.editContent(start, end, change.text);
+      } else {
+        // New text is considered to be the full content of the document.
+        scriptInfo.editContent(0, change.text.length, change.text);
```
kyliau: I think the more appropriate method to call in this case is 
```ts
scriptInfo.open(change.text);
```
This is because we need to replace the entire script snapshot with the new content, and not just a portion of the existing snapshot with the new text.
If we were to stick with `scriptInfo.editContent`, the right way to do this should be
```ts
scriptInfo.editContent(0, scriptInfo.getSnapshot().getLength(), change.text);
```
i.e. we want to replace the *entire* snapshot with the new text.

ayazhafiz: You're right. Thanks!

kyliau: It looks like there wasn't any change, did you push your changes?

Issue #571, opened 2020-01-23
Textmate grammar for template expressions


ayazhafiz:
__Is your feature request related to a problem? Please describe.__

The language service extension now provides a Textmate grammar for the Angular template syntax (#483), and describes template expressions via a JavaScript grammar. However, template expressions are not JavaScript, and to have an accurate grammar long-term, it would be useful to define a grammar specifically for Angular template expressions.

__Additional context__

_Originally posted by @dannymcgee in https://github.com/angular/vscode-ng-language-service/issues/541#issuecomment-577494343_:

@ayazhafiz I've been thinking about the issue with embedded grammars getting stuck and the potential solution I mentioned earlier in this thread, and I made a really rough proof of concept that seems promising.

The basic idea is:
- We embed our own subset of the JavaScript grammar instead of the full JS grammar, limited to just the syntax that's actually valid in Angular templates
- We have two _versions_ of that grammar, one which recognizes single quotes as an end condition, and one that recognizes double quotes
- For each template binding, we inject the correct grammar depending on whether the value was bound with double or single quotes

The hard (or annoying) part of that equation is having to maintain two nearly identical grammars for the JavaScript subset. The solution I came up with was to generate the JSON files dynamically with TypeScript, where each grammar "repository" is defined in a separate file. Then you can just swap out the imports for the handful of repositories that actually need to be different between the two versions. The proof of concept for that is here: https://github.com/dannymcgee/vscode-grammar-from-ts-experiment

There are some other advantages to doing it this way, namely much better code organization and much more readable regex patterns since they can be written as regex literals and benefit from syntax highlighting. And the additional benefits of defining our own grammar for the scripting language is that we can add more accurate grammar scopes for things like pipes and the `as local-val` syntax.

Do you think this is a reasonable solution? And if so, any guidance on how to split up my PRs for ease of review?

_Response; originally posted by @ayazhafiz in https://github.com/angular/vscode-ng-language-service/issues/541#issuecomment-577500730_

@dannymcgee thanks for taking the time to do this. I like your idea of a template syntax grammar; it's definitely the "correct" way to do it.

[...]

PR #581, opened 2020-01-25
refactor: build template.json grammar from typescript source files

dannymcgee:
As discussed in #571:

- Added ts-node as a dev dependency
- Added TypeScript build system for generating JSON gramar files from more legible/organized source code
- Added npm script for building grammars (`yarn build:syntaxes`)
- Refactored Template grammar to TypeScript
- Rebuilt template.json from source


ayazhafiz:
Thanks for the PR! I will take a look tonight; apologies for the tardiness.

dannymcgee:
> Thanks for the PR! I will take a look tonight; apologies for the tardiness.

No worries, and no rush!

syntaxes/src/build.ts
```diff
@@ -0,0 +1,45 @@
...
+  const processedGrammar: JsonObject = {};
+
+  for (const [key, value] of Object.entries(grammar)) {
+    processedGrammar[key] = processValue(value);
+  }
+
+  return processedGrammar;
+};
+
+const build = (grammar: GrammarDefinition, filename: string): void => {
```
ayazhafiz: Please use `function` declarations here rather than arrow functions (they're hoisted, no object scope here, etc).

package.json
```diff
@@ -116,6 +117,7 @@
     "prettier": "1.19.1",
     "rollup": "1.29.1",
     "rollup-plugin-commonjs": "10.1.0",
+    "ts-node": "^8.6.2",
```
ayazhafiz: I am weary of adding a dependency for `ts-node`; I think it would be simpler to compile with tsc and run with node than adding an additional pipeline..

syntaxes/src/build.ts
```diff
@@ -0,0 +1,45 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as fs from 'fs';
+
+import {GrammarDefinition, GrammarDefinitionValue, JsonObject, JsonObjectValue} from './types';
+
+import {template} from './template';
+
+const processValue = (value: GrammarDefinitionValue): JsonObjectValue => {
```
ayazhafiz: Why not overload the function definition here rather than creating `GrammarDefinitionValue` in an external file?

syntaxes/src/build.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (typeof value === 'string') {
+    return value;
+  } else if (value instanceof RegExp) {
+    return value.toString().replace(/^\/|\/$/g, '');
```
ayazhafiz: why is this done?

syntaxes/src/template/index.ts
```diff
@@ -0,0 +1,36 @@
...
+    {include: '#twoWayBinding'},
+    {include: '#templateBinding'},
+  ],
+  repository: {
+    interpolation,
+    propertyBinding,
+    eventBinding,
+    twoWayBinding,
+    templateBinding,
+    bindingKey,
```
ayazhafiz: I would prefer to place all of these definitions in this file, since they are exclusive to the template grammar and very similar, making locality easier to browser. What do you think?

syntaxes/src/types.ts
```diff
@@ -0,0 +1,19 @@
...
+export type GrammarDefinitionValue = string|RegExp|GrammarDefinition|GrammarDefinition[];
+
+export interface GrammarDefinition {
+  [key: string]: GrammarDefinitionValue;
+}
+
+export type JsonObjectValue = string|JsonObject|JsonObject[];
+
+export interface JsonObject {
+  [key: string]: JsonObjectValue;
```
ayazhafiz: Consider

```suggestion
  [key: string]: string|JsonObject|JsonObject[];
```

and elision of the type on line 15

syntaxes/src/build.ts
dannymcgee: - `RegExp.prototype.toString` will do the [escaping](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/toString#Empty_regular_expressions_and_escaping) we need in the output file, while feeding the regex literals directly to `JSON.stringify` will not
- `RegExp.prototype.toString` retains the demarcating `/` characters, so they need to be trimmed from the final output

ayazhafiz: Sorry, I meant to ask why the replacement?

syntaxes/src/template/index.ts
dannymcgee: I personally prefer shorter and separated files because I can open them in multiple panes and look at them side-by-side vs having to scroll all over the place. Especially if/when the grammar becomes more complex, I think this will be more sustainable in the long run vs having a single file that's thousands of lines long, but I suppose that's speculative at this point. It looks like there's no guidance in the Google style guide about file lengths (although [Angular's style guide](https://angular.io/guide/styleguide#rule-of-one) recommends a 400-line limit).

I'll let you consider that and make a final decision one way or the other. :) It's not going to break my heart to consolidate, but that would be my argument for leaving it as-is.

syntaxes/src/build.ts
dannymcgee: The main reason I created the type unions was to avoid having to break these onto multiple lines and to keep the call signatures a little tidier. It also, I think, makes the meaning a little easier to infer (we're transforming values that are valid in a grammar definition into values that are valid in a JSON object, which is maybe a little easier to grok than trying to discern the purpose of transforming `string|RegExp|GrammarDefinition|GrammarDefinition[]` into `string|JsonObject|JsonObject[]`). But abstracting it also kind of obfuscates it. I'm okay with changing it.

syntaxes/src/build.ts
dannymcgee: Fixed!

package.json
dannymcgee: Fixed!

syntaxes/src/build.ts
dannymcgee: Fixed!

syntaxes/src/build.ts
dannymcgee: Added an inline comment to clarify this in the code

syntaxes/src/types.ts
dannymcgee: Fixed!

syntaxes/src/template/index.ts
ayazhafiz: I think we should consolidate in one file because

- symmetry with the outputted JSON file
- the template grammar is a small superset of the HTML grammar, and probably won't be expanded much more than it is now (because we won't rewrite the HTML grammar)

dannymcgee:
> If you write to `syntaxes/out`, I believe you will need to change the grammar contribution file in package.json as well.

The grammars are actually copied from syntaxes/out to dist/syntaxes as part of the build script, per ad747b9894b080dcc920fa6ce99873f357b6ab7c

syntaxes/src/template/index.ts
dannymcgee: @ayazhafiz Sorry for taking so long to get to this, long day at work. Updated with a10718c!

syntaxes/src/build.ts
```diff
@@ -0,0 +1,43 @@
...
+
+// Recursively transforms a TypeScript grammar definition into an object which can be processed by
+// JSON.stringify to generate a valid TextMate JSON grammar definition
+function processGrammar(grammar: GrammarDefinition): JsonObject {
+  const processedGrammar: JsonObject = {};
+  for (const [key, value] of Object.entries(grammar)) {
+    if (typeof value === 'string') {
+      processedGrammar[key] = value;
+    } else if (value instanceof RegExp) {
+      // Escape backslashes/quote marks and trim the demarcating `/` characters from a regex literal
```
ayazhafiz: oh, i see. nice!

PR #584, opened 2020-01-27
Make logging options affect the consoleLogger

mickaelistria:
Most logging messages are not so interesting for end-users and are more
about integration of the LS in the client.
To allow to hide them in clients, make the logging option used by the
session so setting logger to off shouldn't log anything in console.

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -40,11 +40,13 @@ export class Session {
   private readonly projectService: ProjectService;
   private diagnosticsTimeout: NodeJS.Timeout|null = null;
   private isProjectLoading = false;
+  private readonly logger: Logger;
 
   constructor(options: SessionOptions) {
     // Create a connection for the server. The connection uses Node's IPC as a transport.
     this.connection = lsp.createConnection();
     this.addProtocolHandlers(this.connection);
+    this.logger = options.logger,
     this.projectService = new ProjectService({
       host: options.host,
       logger: options.logger,
@@ -385,7 +387,9 @@ export class Session {
    * @param message The message to show.
    */
   error(message: string): void {
-    this.connection.console.error(message);
+    if (this.logger.hasLevel(ts.server.LogLevel.terse)) {
+      this.connection.console.error(message);
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


kyliau:
@mickaelistria Is the goal here to disable logging (to both the remote console and file) completely?

If so, I think there is a bug in the server that prevents LSP clients from disabling logging completely.

According to our help message, the way to completely disable logging is by omitting a log file in the command line args:
https://github.com/angular/vscode-ng-language-service/blob/3a630f2a985d5cc8c5b0874e78b2d8f046da2383/server/src/cmdline_utils.ts#L50-L57

However, if a log file is not provided, the server will redirect the output to stderr instead:
https://github.com/angular/vscode-ng-language-service/blob/3a630f2a985d5cc8c5b0874e78b2d8f046da2383/server/src/logger.ts#L58-L60

This seems incorrect to me. If your intent is to disable logging, I think we should add a flag `off` to `logVerbosity` to address the broader underlying issue.

mickaelistria:
The issue we have is that we see in the user console of Wild Web Developer some messages from Angular Language Server, which are in the vast majority of cases causing noise and confusion to users.
Even with `--logVerbosity terse --logFile /dev/null`, the language server sends messages like
```json
{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":3,"message":"Angular language server process ID: 317715"}}

{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":3,"message":"Using typescript/lib/tsserverlibrary v3.9.7 from /home/mistria/git/wildwebdeveloper/org.eclipse.wildwebdeveloper/node_modules/typescript/lib/tsserverlibrary.js"}}

{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":3,"message":"Using @angular/language-service v10.0.4 from /home/mistria/git/wildwebdeveloper/org.eclipse.wildwebdeveloper/node_modules/@angular/language-service/bundles/language-service.umd.js"}}

{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":3,"message":"Log file: undefined"}}
```

The reason is that those messages are sent via `session.info`, which then delegate to `connection.console.info` which then sends a LS logMessage notification.
We need a way to optionally cut this invocation chain. My proposal is to take the logger options (set via command-line) into account.

kyliau:
@mickaelistria I'm sorry, I dropped the ball on this PR. I think it is better to separate logging to console vs logging to file, and the option `--logVerbosity` is used for logging to file. I suggest we create another flag `--logtoConsole`, and with this flag you can turn off console logging completely. Does that satisfy your use case? See PR #983

PR #592, opened 2020-01-29
refactor: move syntax grammars to TS definitions

ayazhafiz:
This commit refactors remaining existing grammars to use TS definitions
as started in #581.


ayazhafiz:
@dannymcgee if you'd like to take a look

ayazhafiz:
That sounds good. Do you mind if I put that in the next PR? It goes well with putting the generated files in the root syntaxes directory, which makes development easier as well.

kyliau:
Yeah separate PR is fine, thank you!

dannymcgee:
> @dannymcgee if you'd like to take a look

Sorry I'm just seeing this, but looks good to me! The only thing I would note for the next PR is that if we're going to use PascalCase for the grammar identifiers we should probably also rename `template` to `Template`.

PR #604, opened 2020-01-31
build: check in syntax files

kyliau:
This commit makes a few changes:

1. Check in syntax JSON files to syntaxes/ directory
2. Fix lint errors in syntax TS source
3. Add CI check to make sure JSON files are in sync with TS source
4. Add a README


kyliau:
@ayazhafiz PTAL

Issue #575, opened 2020-01-24
Template color highlighting broke


BO41:
**Describe the bug**

Tag highlighting is strange and inconsistent

**Expected behavior**

opening tags and closing tags should have the same color.
No error red should be shown if code is correct.

**Screenshots**

![image](https://user-images.githubusercontent.com/29450609/73054110-79991400-3e81-11ea-92db-389e08404ae4.png)

**Additional context**

worked in `v0.900.5`, broken in `v0.900.6`

smnbbrv:
In addition to the above, the comment shortcut such as `ctrl + /` / `cmd + /` produces a typescript-like comment `// ...` instead of html-like comment `<!-- ... -->`

The issue is really annoying and makes the extension instead of being helpful to absolutely unusable.

My guess: this is somehow related to custom structural directives such as e.g. angular material ones and / or `(event)` handlers.

ghaschel:
> In addition to the above, the comment shortcut such as `ctrl + /` / `cmd + /` produces a typescript-like comment `// ...` instead of html-like comment `<!-- ... -->`
> 
> The issue is really annoying and makes the extension instead of being helpful to be absolutely unusable.
> 
> My guess: this is somehow related to custom structural directives such as e.g. angular material ones and / or `(event)` handlers.
> 
> Here are the versions
> 
> ```
>      _                      _                 ____ _     ___
>     / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
>    / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
>   / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
>  /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
>                 |___/
>     
> 
> Angular CLI: 8.3.23
> Node: 13.6.0
> OS: darwin x64
> Angular: 8.2.14
> ... animations, common, compiler, compiler-cli, core, forms
> ... language-service, platform-browser, platform-browser-dynamic
> ... router
> 
> Package                            Version
> ------------------------------------------------------------
> @angular-devkit/architect          0.803.23
> @angular-devkit/build-angular      0.803.23
> @angular-devkit/build-optimizer    0.803.23
> @angular-devkit/build-webpack      0.803.23
> @angular-devkit/core               8.3.23
> @angular-devkit/schematics         8.3.23
> @angular/cdk                       8.2.3
[...]

PR #609, opened 2020-02-03
Added template expression support

ghaschel:
I changed all the references in template.json from source.js to expression.ng. I had to keep some of the regexes as strings for now.


ghaschel:
@googlebot I signed it!

ayazhafiz:
@ghaschel can you rebase on master?

syntaxes/inline-template.json
```diff
@@ -66,9 +66,6 @@
       },
       "contentName": "text.html",
       "patterns": [
-        {
-          "include": "text.html.derivative"
```
ayazhafiz: Why is this removed?

syntaxes/src/build.ts
```diff
@@ -5,11 +5,12 @@
...
  * found in the LICENSE file at https://angular.io/license
  */
-
 import * as fs from 'fs';
 
-import {GrammarDefinition, JsonObject} from './types';
-import {template} from './template/grammar';
+import { template } from './template/grammar';
+import { expression } from './expression/grammar';
+import { GrammarDefinition, JsonObject } from './types';
```
ayazhafiz: I think you will have to run the formatter (`yarn format`); the CI style check will probably fail on this file.

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
+import { GrammarDefinition } from '../types';
+
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
```
ayazhafiz: Can you place the import statement below this line? The header should be at the very top

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+export const expression: GrammarDefinition = {
+  scopeName: 'expression.ng',
+  injectionSelector: 'L:text.html -comment',
+  patterns: [{ include: '#ngExpression' }],
+  repository: {
+    ngExpression: {
+      name: 'meta.expression.ng',
+      patterns: [
+        {
+          include: '#tsString',
```
ayazhafiz:
```suggestion
          include: '#string',
```

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+        },
+        {
+          include: '#literal',
+        },
+        {
+          include: '#ternaryExpression',
+        },
+        {
+          include: '#expressionOperators',
```
ayazhafiz:
```suggestion
          include: '#expressionOperator',
```

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+          name: 'constant.language.boolean.false.ts',
+          match: /(?<!\.|\$)\bfalse\b(?!\$)/,
+        },
+      ],
+    },
+
+    expressionOperators: {
+      patterns: [
+        {
+          match: /((?<!\|)(\|)(?!\|))\s?([a-zA-Z0-9\-\_]*)/,
```
ayazhafiz: I am confused by what this is matching - a `|` lookbehind, then a  `|`, then another `|` via a lookahead? Can you provide an example where this is the case? Please correct me if I am misinterpreting the pattern.

ayazhafiz:
```suggestion
          match: /((?<!\|)(\|)(?!\|))\s?([a-zA-Z0-9\-\_\$]*)/,
```

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+              name: 'keyword.operator.logical.ts',
+            },
+            '3': {
+              name: 'entity.name.function.pipe.ng',
+            },
+          },
+        },
+        {
+          name: 'storage.type.ts',
+          match: /(?<!\.|\$)\b(let)\b(?!\$)/,
```
ayazhafiz: Similar to above, what is the purpose of matching a literal `$` via lookahead?

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+        {
+          name: 'keyword.operator.expression.instanceof.ts',
+          match: /(?<!\.|\$)\binstanceof\b(?!\$)/,
+        },
+        {
+          name: 'keyword.operator.new.ts',
+          match: /(?<!\.|\$)\bnew\b(?!\$)/,
+        },
+        {
+          include: '#typeofOperator',
```
ayazhafiz: Consider putting this at the bottom or top of all the patterns included in `expressionOperators`

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+        },
+        {
+          name: 'entity.name.function.ts',
+          match: /([_$[:alpha:]][_$[:alnum:]]*)/,
+        },
+        {
+          name: 'meta.type.parameters.ts',
+          begin: /\</,
+          beginCaptures: {
+            '0': {
```
ayazhafiz: You can remove the string from the 0

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+    },
+
+    identifiers: {
+      patterns: [
+        {
+          name: 'support.class.ts',
+          match: /([_$[:alpha:]][_$[:alnum:]]*)(?=\s*\.\s*prototype\b(?!\$))/,
+        },
+        {
+          match: '(?x)(\\?\\.|\\!\\.|\\.)\\s*(?:\n  ([[:upper:]][_$[:digit:][:upper:]]*) |\n  ([_$[:alpha:]][_$[:alnum:]]*)\n)(?=\\s*\\.\\s*[_$[:alpha:]][_$[:alnum:]]*)',
```
ayazhafiz: Is the two spaces intentional?

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+            '1': {
+              name: 'keyword.operator.arithmetic.ts',
+            },
+          },
+        },
+      ],
+    },
+
+    functionCall: {
+      begin: /(?=(\??\.\s*)?([_$[:alpha:]][_$[:alnum:]]*)\s*(<([^<>]|\<[^<>]+\>)+>\s*)?\()/,
```
ayazhafiz: Will `[:alpha:]` match all alpha characters or only :, a, l, p, h?

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+        },
+      ],
+    },
+
+    thisLiteral: {
+      name: 'variable.language.this.ts',
+      match: /(?<!\.|\$)\bthis\b(?!\$)/,
+    },
+
+    tsString: {
```
ayazhafiz:
```suggestion
    string: {
```

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+      patterns: [
+        {
+          include: '#qstringSingle',
+        },
+        {
+          include: '#qstringDouble',
+        },
```
ayazhafiz: What about template strings ("`")?

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+          end: /(?<=\))/,
+          patterns: [
+            {
+              include: '#functionParameters',
+            },
+          ],
+        },
+        {
+          name: 'meta.type.function.ts',
+          begin: '(?x)( (?= [(]\\s*( ([)]) |  (\\.\\.\\.) | ([_$[:alnum:]]+\\s*( ([:,?=])| ([)]\\s*=>) )) ) ) )',
```
ayazhafiz: Same as some previous comment. Are the spaces necessary? The string is okay; we can fix that up later.

syntaxes/test/dummy/expression.ng.tmLanguage-dummy.json
```diff
@@ -0,0 +1,4 @@
+{
+  "comment": "Dummy Angular expression TextMate grammar for use in testing",
+  "scopeName": "expression.ng"
+}
```
ayazhafiz: I don't think we want this... we should test against the new grammar you provided :).

syntaxes/test/cases.json
```diff
@@ -14,6 +14,12 @@
     "grammarFiles": ["syntaxes/inline-styles.json"],
     "testFile": "syntaxes/test/data/inline-styles.ts"
   },
+  {
+    "name": "expression syntax",
+    "scopeName": "expression.ng",
+    "grammarFiles": ["syntaxes/out/expression.json"],
```
ayazhafiz: The tests look a little off right now, because they all point to `expression.ng`, but do not show any of the new scopes you have created in the expression grammar. You will probably have to delete the dummy grammar and add `syntaxes/template.json` to the grammar files here.

syntaxes/test/data/expression.html
```diff
@@ -0,0 +1,63 @@
...
+
+<div [@animation.trigger]="val"></div>
+<img [attr.aria-label]="val" />
+<div [my-property]="val"></div>
+<div [my_property]="val"></div>
+<div [myProperty$]="val"></div>
+<div [%invalidProperty]="val"></div>
+<div [invalidProperty)="val"></div>
+
+<!-- Event binding test -->
```
ayazhafiz: Can you refactor this test file to more appropriately test the various scopes provided by your grammar? The expression grammar test probably doesn't care too much that the event binding is picked up probably, because that is covered by the template grammar test. Here we should test that the different strings, expression operators, etc. work correctly.

syntaxes/src/expression/grammar.ts
dannymcgee: @ayazhafiz Are template literals valid in template expressions? I didn't think they were but I could be wrong

syntaxes/src/expression/grammar.ts
ghaschel: Most of this grammar I have it derived from [this file](https://github.com/microsoft/vscode-textmate/blob/master/test-cases/themes/syntaxes/TypeScript.tmLanguage.json#L1805) from the microsoft vscode-textmate repository. I don't think the spaces will make any difference to be honest. I'll remove it, and we can later restore it if needed

dannymcgee: I think these patterns beginning with the ignore-whitespace flag `(?x)` are a holdover from when they were converted from the XML format (in that format, patterns with this flag [can be split onto multiple lines](https://github.com/Microsoft/TypeScript-TmLanguage/blob/master/TypeScript.tmLanguage) which makes them a little more legible).

In my experience, the way to successfully refactor these is to remove the `(?x)` flag, then remove all whitespace characters (including `\n`s) and comments (beginning with `#` and usually ending with a bunch of whitespace characters) from the pattern.

From there, refactoring them to regex literal should just be a matter of replacing the demarcating quotes with `/` and removing the redundant escaping backslashes. We could also replace the Oniguruma POSIX brackets with their equivalent character classes for consistency and better TypeScript syntax highlighting (e.g. `[_$[:alnum:]]` becomes `[_$a-zA-Z0-9]` -- more info [here](https://github.com/stedolan/jq/wiki/Docs-for-Oniguruma-Regular-Expressions-(RE.txt))), but that's not super important IMO.

syntaxes/src/expression/grammar.ts
ghaschel: Angular does not allow lexer characters in the template expression as far as I know... IMO it's up to the template grammar to match them

syntaxes/src/expression/grammar.ts
ghaschel: I agree with replacing the Oniguruma POXIX brackets as they are not that intuitive. I think that we can hold that a bit because there are a few more of these

syntaxes/src/expression/grammar.ts
```diff
@@ -0,0 +1,886 @@
...
+      ],
+    },
+
+    arrayLiteral: {
+      name: 'meta.array.literal.ts',
+      begin: /\[/,
+      beginCaptures: {
+        '0': {
+          name: 'meta.brace.square.ts',
+        },
```
dannymcgee: Sorry for the nitpick -- would you be willing to remove the quotes from the numeric keys and put the objects with only one key/value pair on a single line for consistency with the Template grammar? I think it makes it a little more readable, and consistency is nice -- but it's also kind of tedious, so I wouldn't blame you for leaving it as-is for now. :)

ghaschel: I'll sure do it. I missed all of them. Got used to using prettier and completely forgot about it, it does all the dirty work. haha

syntaxes/src/expression/grammar.ts
ghaschel: This is residual from the [base file](https://github.com/microsoft/vscode-textmate/blob/8961b1554acbb649bdce0d236ab0fd78180447af/test-cases/themes/syntaxes/TypeScript.tmLanguage.json#L2284). I'll remove it.

dannymcgee: This first group could be simplified to this: `([?!]?\\.)` which makes it a little easier to parse. (Not sure if the same pattern shows up elsewhere in the grammar; just noticed it while I was scrolling through the source)

syntaxes/src/expression/grammar.ts
ghaschel: `[:alpha:]` is a POXIS bracket used in Oniguruma, the vscode regex engine. This is the same as `[a-zA-Z]`

syntaxes/src/expression/grammar.ts
ghaschel: I kept this in `let` to keep the pattern that was already being used in the base file. I believe the negative lookahead and lookbehind are basically preventing the scope `storage.type.ts` to being added to `.let`, `$let` and `let$`. A trailing `$` is often used as a variable naming convention for observables.

syntaxes/src/expression/grammar.ts
ghaschel: This is used to match pipes.

I am matching a `|` that doesn't have a `|` before and another after, and are followed by a alphanumeric characters.

I made it this way to prevent it from interacting with the `||` operator, as it is allowed in the same level as the pipes.

syntaxes/inline-template.json
ghaschel: This was an error of mine. Sorry.

syntaxes/src/expression/grammar.ts
ghaschel: Done

syntaxes/src/expression/grammar.ts
ghaschel: changed all of the references from tsString to string

ghaschel: I'm skipping commiting the suggestion here as there are the references to the pattern in the file

syntaxes/src/expression/grammar.ts
ghaschel: Changed the references and skipping commiting here as there are references to the pattern in the file aswell

syntaxes/src/expression/grammar.ts
ghaschel: Removed from all of the numbers

syntaxes/src/expression/grammar.ts
ghaschel: Removed from all of the numbers

syntaxes/src/expression/grammar.ts
ghaschel: Changed to the end of the array

syntaxes/src/expression/grammar.ts
ayazhafiz: okay good call

syntaxes/src/expression/grammar.ts
ayazhafiz: This seems like a pretty specific match. I think we could instead demarcate statements by pipe and semicolon, and then make the `||` operator a level below as something allowed as an expression in the statement. What do you think?

syntaxes/src/expression/grammar.ts
ayazhafiz: Okay, I see. Wow, they really are covering all the cases..

syntaxes/src/expression/grammar.ts
ayazhafiz: wow. cool.

syntaxes/src/expression/grammar.ts
ghaschel: The problem with demarcating statements by semicolon in this case is that it is not mandatory, and often are removed automatically by code formatters. 

The only way I found to make pipes work properly was inserting in at the first level of the expression pattern alongside all of the operators.

It is now pretty solid and it won't conflict with the `||` operator because it is first in the pattern array, so it gets the top priority. And also the matching itself excludes each other.

I won't object in finding another way to do it, but IMO, I don't think it is that urgent.

syntaxes/src/expression/grammar.ts
ghaschel: I changed this, thank you for pointing it out.

syntaxes/src/expression/grammar.ts
ghaschel: I removed the spaces here too.

syntaxes/src/expression/grammar.ts
dannymcgee: A potential simplification would be to replace the negative lookahead with a `{1}` and replace the `*` in the second group with `+` (since you can't have a pipe with no identifier) — but you do still need the negative lookbehind so it's honestly not a huge improvement.

One thing I would recommend is removing the extra parens around the `\|` in the middle of the first group, because that creates a duplicate capture group (compare the Details tab [here](https://regexr.com/4tk7o) and [here](https://regexr.com/4tk7r) with the match selected)

syntaxes/src/expression.ts
```diff
@@ -0,0 +1,892 @@
...
+              name: 'keyword.control.as.ts',
+            },
+          },
+          end: /(?=$|"|[;,:})\]])/,
+          patterns: [
+            {
+              include: '#type',
+            },
+          ],
+        },
```
dannymcgee: Sorry for the nitpick — is there any way we could scope this as a variable instead of a type? The `as` keyword in template expressions isn't a typecast, it's an assignment of the preceding expression to a local variable. It's always bugged me that the JS/TS grammar colors it as a type. :P But it's really not a huge deal, and if need be I can open a separate PR for that change.

syntaxes/src/expression.ts
```diff
@@ -0,0 +1,892 @@
...
+          match: /((?<!\|)(\|)(?!\|))\s?([a-zA-Z0-9\-\_\$]*)/,
+          captures: {
+            2: {
+              name: 'keyword.operator.logical.ts',
+            },
+            3: {
+              name: 'entity.name.function.pipe.ng',
+            },
+          },
+        },
```
dannymcgee: If you remove the parens around the `\|` in the middle of the first group, it won't create a duplicate group and then you can change the `captures` keys to `1` and `2` instead of `2` and `3`. Compare the Details tab [here](https://regexr.com/4tk7o) and [here](https://regexr.com/4tk7r) with the match selected.

syntaxes/src/expression.ts
ghaschel: IMO I think having `index as i` having all of the 3 the same color is something that doesn't make it any easier to read... But if you guys think it fits better, I'm up to that. @ayazhafiz what do you think of this?

syntaxes/src/expression.ts
ghaschel: Removing parenthesis from the negative lookeahead and negative lookbehind invalidate it. (At least in [Rubular](https://rubular.com/) - I use Rubular to test my regexes because it is made using Ruby, and Oniuruma is the Ruby regex engine).

But removing the enclosing parenthesis from the lookbehind, `|` capture group and lookahead makes it capture only once.

`((?<!\|)(\|)(?!\|))\s?([a-zA-Z0-9\-\_\$]*)` became  `(?<!\|)(\|)(?!\|)\s?([a-zA-Z0-9\-\_\$]*)`

ghaschel: Good call! Done that

syntaxes/src/expression/grammar.ts
ghaschel: @dannymcgee this is not resolved yet, but It's fine creating a separate comment

syntaxes/src/expression.ts
dannymcgee: > IMO I think having `index as i` having all of the 3 the same color is something that doesn't make it any easier to read...

The `as` there would still be a keyword, so it would only be `index` and `i` that get the same color (sorry if that wasn't clear). But FWIW, the more common pattern I see in the wild is `*ngIf="(foo$ | async) as foo"`, and with the `async` there being colored as a function there's still plenty of differentiation. 

Either way, `index as i` is functionally equivalent to `let i = index`, so it strikes me as weirdly inconsistent that the variable name would be colored differently depending on the syntax used to make the assignment.

ghaschel: Oh, I understand now. From that perspective, the adequate would be having `let` and `as` having the same color, as they do fundamentally the same thing.

In that case, `as` would be in `storage` instead of `keyword`

ghaschel: image

syntaxes/test/data/expression.html
ghaschel: Sure. I'll do it tonight

syntaxes/src/expression.ts
ayazhafiz: I think this is reasonable. Longer term we should spec out a formal grammar for this.

dannymcgee: > In that case, `as` would be in `storage` instead of `keyword` alongside with `let`

I think VS Code's theme colors all keywords dark blue with the exception of control keywords which get the purple. It's definitely not a control keyword; storage sounds reasonable to me.

But the thing that actually bugs me here is that the identifier _after_ the `as` is colored as a type (green in VS Code's theme) instead of a variable (light blue). My theme colors types with an italicized orange and local variables with a light pink, so the difference is pretty jarring. :P

But I'm also fine with holding off on this until another PR and maybe taking some time to spec out the grammar in the meantime.

syntaxes/test/data/expression.html
```diff
@@ -0,0 +1,66 @@
+<!-- Interpolation test -->
```
ayazhafiz: Let's just put each test case in an interpolation like `{{ }}`, each on a separate line and testing a clear, individual thing. Some "integration" tests for combinations of grammar scopes can be included at the end too. But the tags and attributes are fluff that we test in other files, and I think here it is much more important to test that we handle operators, variable assignments, pipes, etc. correctly.

If possible, can you include one test case for every scope you define in your grammar and some integrations between them at the end? I realize this is a big task; if you would like some help, please let me know and I will work on it as well.

syntaxes/src/expression.ts
ayazhafiz: I think it will be nearly impossible to solve all open issues in this PR, given its existing size. Though I would have preferred incremental additions of the grammar, which would probably elide these quirks, a lot of time has already been put into this PR already. Given this, I think as long as the grammar testing coverage is thorough and looks reasonable, we can go ahead and merge and apply hotfixes as well as a formal grammar later on. We should stay from additions more than a few hundred lines, let alone nearly 3000, moving forward.

syntaxes/test/data/expression.html
ghaschel: Sorry for taking so long to answer. I had to work off shift yesterday.

I think the best option for now would be covering the main expression cases (all of the common expression cases). All of the scopes wont be possible now as there are some unused scopes still present in the code. After having this PR merged, I'll focus on reducing a bit more the grammar to clear up those unused scopes.

Does that sounds good?

BTW: When working on reducing this grammar, I'll eventually be adding these changes to the grammar in my extension as well. Would you mind if I "borrow" the way this extension approaches snapshot tests? I find it very productive.

ghaschel: I'll be adding test cases to the html file and as soon as you find it to be enough, I'll rebase on master and I think we're ready to go.

ayazhafiz: Sure, that all sounds good. Feel free to reuse as much of anything as you would like.

Following this PR we should talk about a formal grammar for expressions. There are actually two things going on here: a grammar for expressions as they exist e.g. inside interpolations, and the microsyntax that we have for template bindings. I'll present a doc later on.

syntaxes/src/build.ts
```diff
@@ -5,9 +5,9 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-
```
ayazhafiz: nit: can you add this back

syntaxes/test/cases.json
```diff
@@ -11,6 +11,12 @@
...
     "testFile": "syntaxes/test/data/inline-styles.ts"
   },
+  {
+    "name": "expression syntax",
+    "scopeName": "expression.ng",
+    "grammarFiles": ["syntaxes/expression.json"],
+    "testFile": "syntaxes/test/data/expression.html"
+  },
   {
     "name": "template syntax",
```
ayazhafiz: Let's load the expression syntax on this test case as well.

syntaxes/test/data/expression.html
```diff
@@ -0,0 +1,80 @@
...
+{{ call() }}
+{{ call(); }}
+{{ call().object }}
+{{ call().object; }}
+{{ call()?.conditional }}
+{{ call()?.conditional; }}
+{{ call()!.conditional }}
+{{ call()!.conditional; }}
+
+<!-- Nested function call -->
```
ayazhafiz:
```suggestion
<!-- Received property read and method call -->
```

This is the language used in the compiler..

syntaxes/test/data/expression.html
```diff
@@ -0,0 +1,80 @@
...
+{{ {object: null} }}
+{{ {object: this} }}
+{{ {object: undefined} }}
+{{ {object: [variable, variable]} }}
+{{ {object: ['string', 'string']} }}
+{{ {object: ["string", "string"]} }}
+{{ {object: [1, 2]} }}
+{{ {object: [true, false, null, this, undefined]} }}
+
+<!-- Function call -->
```
ayazhafiz:
```suggestion
<!-- Function call and property read -->
```

syntaxes/test/data/expression.html
```diff
@@ -0,0 +1,80 @@
...
+{{ condition ? true : false }}
+{{ condition ? true : false; }}
+{{ condition() ? call(1 + 2 + 3) : call()  }}
+{{ condition() ? call(1 + 2 + 3) : call();  }}
+{{ condition()?.object ? call()!.test() : false }}
+{{ condition ? ['123'] : { test: 'a' } }}
+{{ condition ? 'test' : "test" }}
+{{ condition ? [function(), variable] : {} }}
+
+<!-- Expression test -->
```
ayazhafiz:
```suggestion
<!-- Microsyntax test -->
```

These I think should be placed in `ngFor`s or similar.

syntaxes/test/data/expression.html
```diff
@@ -0,0 +1,80 @@
...
+{{ (let param of params?.anotherParam!.param().param | async) as p; let i = index; let first = first }}
+{{ let param of params | async as p; let i = index; let first = first; }}
+{{ (let param of params | async) as p; index as i; first as first; }}
+{{ value | pipe:'with-param' }}
```
ayazhafiz: Let's make pipes their own section

syntaxes/test/data/expression.html
```diff
@@ -0,0 +1,80 @@
...
+{{ condition ? true : false; }}
+{{ condition() ? call(1 + 2 + 3) : call()  }}
+{{ condition() ? call(1 + 2 + 3) : call();  }}
+{{ condition()?.object ? call()!.test() : false }}
+{{ condition ? ['123'] : { test: 'a' } }}
+{{ condition ? 'test' : "test" }}
+{{ condition ? [function(), variable] : {} }}
+
+<!-- Expression test -->
+{{ (let param of params?.anotherParam!.param().param | async) as p; let i = index; let first = first }}
```
ayazhafiz: Let's make this a little more atomic. It's not clear what's being tested here..

syntaxes/test/data/expression.html
```diff
@@ -0,0 +1,80 @@
...
+{{ let param of params | async as p; let i = index; let first = first; }}
+{{ (let param of params | async) as p; index as i; first as first; }}
+{{ value | pipe:'with-param' }}
+{{ { test: 'value'} | json }}
+{{ value && value || value && !!value }}
```
ayazhafiz: Can you add test cases specifically for the issue seen in https://github.com/angular/vscode-ng-language-service/issues/613 and https://github.com/angular/vscode-ng-language-service/issues/575 in the pipes and variable assignment sections, respectively?

syntaxes/test/cases.json
ghaschel: I'm sorry, I don't think I understand what you mean.

ayazhafiz: Could you add the expression grammar to the grammarFiles for the template syntax test case as well?

ayazhafiz: Or attach to it a dummy grammar for expression.ng. If you look at the template snap file it looks like it’s not picking up the expression grammar at all.

ayazhafiz:
@ghaschel could I help you with committing some of these changes? It could speed up the time of the PR.

syntaxes/test/cases.json
ghaschel: > @ghaschel could I help you with committing some of these changes? It could speed up the time of the PR.

I'm sorry for the (very) late response... I'm having the craziest of the weeks at work... I think tomorrow I'll be able to work on this... But any help would be appreciated

ayazhafiz: No worries, man. I totally understand. Sounds good.

syntaxes/test/data/expression.html
ghaschel: That's okay. We can do that. 

Sorry for the long time to asnwer. i've been busier than I expected

syntaxes/test/cases.json
ghaschel: I have added you as a collaborator

syntaxes/test/data/expression.html
ghaschel: Done that. let me know if that looks good

syntaxes/test/data/expression.html
ghaschel: I added a specific part on the tests as "mixed test" to test the microsyntax against the expression

syntaxes/test/data/expression.html
ghaschel: I have added a specific session for each case

syntaxes/src/build.ts
ghaschel: Done

syntaxes/test/data/expression.html
ayazhafiz: No worries. I will work on this today.

ayazhafiz:
@googlebot I consent.

ayazhafiz:
@ghaschel if this looks good to you, I think we can go ahead and merge! 🎉

ghaschel:
@ayazhafiz looks good to me

ghaschel:
@ayazhafiz I rebased on master. Let me know if there's something wrong.

ayazhafiz:
@ghaschel I rebased again on master with your changes. Can you confirm that this is still okay with you and we will merge?

ayazhafiz:
Screens btw:

Pre-PR
Screen Shot 2020-02-19 at 4 42 52 PM

Now
Screen Shot 2020-02-19 at 4 43 55 PM

lincolnthree:
@ayazhafiz You are a HERO.

ayazhafiz:
@ghaschel and @dannymcgee are heroes!

ghaschel:
@ayazhafiz I'm fine with that. I think we are ready to merge.

Issue #632, opened 2020-02-13
Editor buffer out of sync with file on disk


aditbharadwaj:
**Describe the bug**

The Angular Language Service server crashed 5 times in the last 3 minutes. The server will not be restarted. Even after restarting the Vscode the service keeps on crashing.

**To Reproduce**

1. Service Crashing after 5 minutes of working

**Logs**
1. Console Log
[output log of the vscode.txt](https://github.com/angular/vscode-ng-language-service/files/4197743/output.log.of.the.vscode.txt)
2. Log File
[nglangsvc.log](https://github.com/angular/vscode-ng-language-service/files/4197748/nglangsvc.log)

**Screenshots**
![als](https://user-images.githubusercontent.com/6751554/74425914-75598880-4e7a-11ea-913f-c22b8cecfab8.PNG)

ayazhafiz:
I think I know what is causing this, but I cannot reproduce the error. Are you able to show what you were editing when the crash happened?

andrius-pra:
Could you try to remove "tslint-language-service" from your tsconfig.json?

ayazhafiz:
@andrius-pra I loaded a project with 'tslint-language-service', and while the package could not be found, it didn't crash the language service. I could be wrong though.

There are two kinds of errors in the log. One of them is the result of a failed assertion during the editing of a source file (https://github.com/microsoft/TypeScript/blob/950b36192a10f5dac52993f1e29e62cbac513853/src/compiler/parser.ts#L8052). I think this may be caused by how we pass changes to source files: https://github.com/angular/vscode-ng-language-service/blob/fd3dfd92e67b57622955af09985e2d073c6cf336/server/src/session.ts#L233-L239

which changed since #553.

andrius-pra:
@ayazhafiz, I have reproduced this issue. 
![bug](https://user-images.githubusercontent.com/17968027/74551579-1d934e00-4f5c-11ea-986c-da82edf9bd48.gif)

ayazhafiz:
Nice!

aditbharadwaj:
> I think I know what is causing this, but I cannot reproduce the error. Are you able to show what you were editing when the crash happened?

i was actually editing two files in split view and if u want the content i was basically calling variables and checking their definition. that's it. mainly it was whenever i was in split screens with multiple files it crashed for me.

kyliau:
[...]

PR #638, opened 2020-02-14
fix: editor buffer out of sync with file on disk

kyliau:
In https://github.com/angular/vscode-ng-language-service/pull/476 it was
incorrectly assumed that the content on disk is the most up-to-date
version. An unsaved buffer in the editor could be newer than the file on
disk and thus we should always load the script with the snapshot from
editor.

PR fix https://github.com/angular/vscode-ng-language-service/issues/632

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -166,19 +166,17 @@ export class Session {
   }
 
   private onDidOpenTextDocument(params: lsp.DidOpenTextDocumentParams) {
-    const {uri, languageId} = params.textDocument;
+    const {uri, languageId, text} = params.textDocument;
     const filePath = uriToFilePath(uri);
     if (!filePath) {
       return;
     }
     const scriptKind = languageId === LanguageId.TS ? ts.ScriptKind.TS : ts.ScriptKind.External;
-    // Since we are only keeping track of changes of files that are already on
-    // disk (see documentSelector.scheme in extension.ts), the content on disk
-    // is up-to-date when a file is first opened in the editor.
-    // In this case, we should not pass fileContent to projectService.
-    const fileContent = undefined;
     try {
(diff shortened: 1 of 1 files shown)


kyliau:
I don't remember why #476 was needed to remove `getExternalFiles()`...
But from my local testing nothing is broken after #476 was reverted.

server/src/session.ts
```diff
@@ -166,19 +166,17 @@ export class Session {
...
-    // disk (see documentSelector.scheme in extension.ts), the content on disk
-    // is up-to-date when a file is first opened in the editor.
-    // In this case, we should not pass fileContent to projectService.
-    const fileContent = undefined;
     try {
-      const result = this.projectService.openClientFile(filePath, fileContent, scriptKind);
+      // The content could be newer than that on disk. This could be due to
+      // buffer in the user's editor which has not been saved to disk.
+      // See https://github.com/angular/vscode-ng-language-service/issues/632
+      const result = this.projectService.openClientFile(filePath, text, scriptKind);
```
ayazhafiz: Will this only ever be called on files that already exist on disk? The other case is that we have a buffer that has never been recorded on disk:

https://github.com/angular/vscode-ng-language-service/blob/3c7f64b4c3805087b8055644f1a7a01fd68a2211/client/src/extension.ts#L28-L33

and in this case, what will happen if we try to open the client file?

kyliau: If a buffer is not on disk, it's an "untitled" document. Since untitled document is not part of the document selector that we specified, we will never receive the `DidOpenTextDocument` event for such buffer.

ayazhafiz: Okay. I think this is a mixup that caused the change in 476. LGTM

PR #701, opened 2020-03-25
Log message when language service is enabled for a project

ayazhafiz:


server/src/session.ts
```diff
@@ -457,6 +461,13 @@ export class Session {
         this.connection.console.info(msg);
         project.log(msg);
       }
+
+      return;
+    }
+    // The language service should be enabled at this point.
+    if (!this.languageServiceEnabled.get(project)) {
```
kyliau: Do we actually need to do this check?
At this point we already know that the language service is enabled, so we could just go ahead and print the statement.
Is this check to make sure that we don't print an activation message for a particular project twice?

ayazhafiz: Yeah, that was the idea. But I guess it's fine to print it multiple times. Will update

PR #708, opened 2020-03-29
Remove .travis.yml

ayazhafiz:
Since we have moved to Circle CI, there is no reason to keep the
.travis.yml file.


kyliau:
hmm ... CIrcleCI wasn't triggered in this PR. I have to check what's going on ..

Issue #521, opened 2019-12-19
event binding and property binding user experience


doggy8088:
**Describe the bug**

Related to https://github.com/angular/vscode-ng-language-service/issues/474

The developing experience for event binding and property binding user experience is changed in this new version `0.900.3`.

**To Reproduce**

1. find a tag
2. add a `(click)=""` event binding or `[accessKey]=""` property binding

**Expected behavior**

When you type `(cli` then hit `Tab` should appear `(click)="<CURSOR_HERE>"`.

**Logs**
N/A
**Screenshots**

`0.900.3` Video
https://youtu.be/JB23F7vflL8

`0.802.3` Video
https://youtu.be/i1XoRvyibjo

The original (`0.802.3`) DX is a bit nicer. Please change it back to the original behavior.

**Additional context**

Hope to get the same or better developer experience (DX) of the code snippet feature.

ayazhafiz:
@kyliau we can probably extend the replacement span to include the closing parantheses in these cases, since we know know the kind of binding we're in before providing a completion. What do you think?

ayazhafiz:
@doggy8088 does this remain a strong pain point for you?

doggy8088:
@ayazhafiz I'm using the latest version (`0.900.15`). The UX issue still there. Let me give you an example.

1. Pick any tag, add an event binding syntax such as `(click)=""`
2. Enter `(`, enter `cli`, then hit `Tab` to complete. The current version will produce:

    ```ts
    (click|)
    ------↑
    ```

    I expected to get the following result:

    ```ts
    (click)="|"
    ---------↑
    ```

ayazhafiz:
Okay, we will take a look at what can be done here :)

doggy8088:
@kyliau The event binding is broken again in `v12.0.0`. There is no IntelliSense at Angular v12.0.0 version in the Visual Studio Code.

dragonhelios:
can add to this that by changing settings to use legacy view instead of ivy the intellisense seems to be working again

doggy8088:
@ayazhafiz The event binding's IntelliSense is still broken in `12.2.0`.

ivanwonder:
I think we can pick the PR #820 up, move the code into `@angular/language-service`, and this feature can be opt-out via the config [`includeCompletionsWithSnippetText`](https://sourcegraph.com/github.com/microsoft/TypeScript/-/blob/lib/tsserverlibrary.d.ts?L9415:18).
@atscott

PR #820, opened 2020-06-15
feat: provide snippets for attribute and method completions

ivanwonder:
when the user selects the attribute in the completion list, the position of the cursor is not expected. It should be in the middle of quotes.

### DIFF
--- a/server/src/completion.ts
+++ b/server/src/completion.ts
@@ -77,9 +77,81 @@ export function tsCompletionEntryToLspCompletionItem(
   // Text that actually gets inserted to the document. It could be different
   // from 'entry.name'. For example, a method name could be 'greet', but the
   // insertText is 'greet()'.
-  const insertText = entry.insertText || entry.name;
-  item.textEdit = entry.replacementSpan ?
-      lsp.TextEdit.replace(tsTextSpanToLspRange(scriptInfo, entry.replacementSpan), insertText) :
+  let insertText = entry.insertText || entry.name;
+  let replaceSpan = entry.replacementSpan;
+  if (kind === CompletionKind.attribute) {
+    // <div [(ng|)]=></div>
+    //      ^-------------- start
+    //             ^------- end
+    let start = scriptInfo.lineOffsetToPosition(position.line + 1, position.character) - 1;
+    let end = scriptInfo.lineOffsetToPosition(position.line + 1, position.character) + 4;
+    if (replaceSpan) {
+      start = replaceSpan.start - 2;
+      end = replaceSpan.start + replaceSpan.length + 3;
+    }
+    const attributeText = scriptInfo.getSnapshot().getText(start, end);
+    switch (attributeText[1]) {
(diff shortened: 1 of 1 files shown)


ayazhafiz:
re #521

ivanwonder:
Yes, I think it's possible to move the code to the upstream package. Because this PR only does is rewrite the `insertText` and `replacementSpan`. But I just meet a problem here, for example `(clic|)`,  there will be no provided info for the user in vscode when the `replacementSpan` starts from `(`. https://github.com/angular/vscode-ng-language-service/blob/1ff56baad7ffe971e3ebbef500ed6642a3d84cff/server/src/completion.ts#L108
Maybe I need to look into it.

kyliau:
> Thanks for the fix, this has been requested for some time. However, I think it would be better to provide this in the language service package so that other Angular language servers, like CoC can take advantage of it too. It is easier to test in the upstream package as well. WDYT?

More to your point this feature relies on snippet, which is available in the LSP protocol but not available in the Typescript interface. I think it's a good feature to have, but we should discuss if we want to have such complex logic in the extension layer.

I've briefly read about "Actions", but do not understand enough to know how they are used. Perhaps we could explore it a little more and see if it could be used here.

https://github.com/microsoft/TypeScript/blob/6c11ceb8db0a085144ee0041793b1e7faa4c172f/lib/tsserverlibrary.d.ts#L5752-L5768

ayazhafiz:
My two cents: I don't think @angular/language-service needs to in total alignment with the tsserver API. We aren't currently. In my opinion the language service package should provide the best features it can, and then it should be the responsibility of the extension implementations to convert the features provided by the language service into types expected by the LSP implementation the extension is targeting. Of course the two pieces shouldn't diverge very far, but keeping the features of the core language service in one component makes it easier to develop new features without having to wait for users to catch up.

ivanwonder:
I think it's better to let the `@angular/language-service` know the [client's capability](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion).


> Yes, I think it's possible to move the code to the upstream package. Because this PR only does is rewrite the `insertText` and `replacementSpan`. But I just meet a problem here, for example `(clic|)`, there will be no provided info for the user in vscode when the `replacementSpan` starts from `(`.
> 
> https://github.com/angular/vscode-ng-language-service/blob/1ff56baad7ffe971e3ebbef500ed6642a3d84cff/server/src/completion.ts#L108
> 
> 
> Maybe I need to look into it.

I found that the vscode will filter the completion items by the text cut by `textEdit`.
image
If we provide it in LS, Maybe we should return the completion item label with the symbol, for example `(click)`

ivanwonder:
I find the issue about [snippet for TSServer](https://github.com/microsoft/TypeScript/issues/25207)

atscott:
Closing, since we're implementing this in the @angular/language-service and then adding an option for handling it in the @angular/language-server

PR #863, opened 2020-07-17
refactor: remove project_service.ts

kyliau:
`ProjectService` class in `project_service.ts` is just a thin wrapper around
`ts.server.ProjectService`. It was added primarily to override the default
behavior of `projectService.getDefaultProjectForFile()` and to prevent
accidental calling of the original method.

We have to override the method because the `ScriptInfo` for a HTML file
does not necessarily have a default project. In the overridden method, we
manually attach it to a project based on the closest `tsconfig.json`.

However, this wrapper complicates the migration to Ivy because when it
configures the plugin, it assumes the name of the plugin is always
`@angular/language-service`, which is not true in the Ivy case.

The methods that were overridden are now absorbed into `session.ts`.


ayazhafiz:
> it assumes the name of the plugin is always
> `@angular/language-service`, which is not true in the Ivy case.

Why is this not true in the ivy case?

kyliau:
> Why is this not true in the ivy case?

tsserver loads the Angular plugin by resolving the name `@angular/language-service`.
Since Ivy is not yet the main entry-point of the package, we'll have to load Ivy LS using 
`@angular/language-service/bundle/ivy`, and that becomes the name of the plugin.

ayazhafiz:
Okay that was my assumption, just wanted to make sure it wasn't something more complicated

PR #864, opened 2020-07-17
fix: run format check iff on CI

kyliau:
When running `yarn format` locally the CI check should not be triggered.

Run the formatting check only when "CIRCLECI" is present in the environment
variables.
`CIRCLECI` is a built-in environment variable that's always set according to
https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

### DIFF
--- a/scripts/format.sh
+++ b/scripts/format.sh
@@ -12,7 +12,7 @@ find . -name "*.ts" \
 find syntaxes/ -name "*.json" \
   -exec yarn prettier --write {} +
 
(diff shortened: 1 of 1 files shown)


kyliau:
This is a regression caused by https://github.com/angular/vscode-ng-language-service/pull/851, but in the original script we checked for `CI_MODE`, which is not a built-in environment variable (perhaps it was used in Travis back then?)

ayazhafiz:
Travis doesn't have CI_MODE, maybe it was some compat thing when we had both travis and circle

PR #865, opened 2020-07-18
feat: add --experimental-ivy flag for Ivy language service

kyliau:
This commit adds a hidden flag `--ivy` to load the Ivy version of
`@angular/language-service`.
This flag is hidden for now because it is only used for development.

The second entry point is resolved from `@angular/language-service/ivy.umd`
but in the next update it'll be `@angular/language-service/ivy` due to
the renaming done in https://github.com/angular/angular/pull/38086

### DIFF
--- a/server/src/cmdline_utils.ts
+++ b/server/src/cmdline_utils.ts
@@ -28,6 +28,11 @@ function hasArgument(argv: string[], argName: string): boolean {
 
 interface CommandLineOptions {
   help: boolean;
+  /**
+   * If true, use the Ivy version of Angular LS. For now this is only used for
+   * development.
+   */
+  ivy: boolean;
   logFile?: string;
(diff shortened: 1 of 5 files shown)


server/src/cmdline_utils.ts
```diff
@@ -37,6 +42,7 @@ interface CommandLineOptions {
 export function parseCommandLine(argv: string[]): CommandLineOptions {
   return {
     help: hasArgument(argv, '--help'),
+    ivy: hasArgument(argv, '--ivy'),
```
ayazhafiz, line 45: nit: I would prefer this be named `--unstable-ivy` or similar, I am worried that someone will run this, expect it to be stable, and file issues.

kyliau: renamed to `--experimental-ivy`

PR #890, opened 2020-08-17
fix: do not use require() to read JSON files

kyliau:
This commit fixes a potential security vulnerability caused by the use of
require() to read a JSON file. If there is a JS file named `package.json.js`
in the node_modules, we would unknowingly execute the remote code while
trying to require() it.

Even though we only load `typescript` and `@angular/language-service`,
we cannot guarantee the content of these packages when they are installed
on users' machines.

Thanks @ddworken for reporting this issue and @koto for surfacing it to the
Angular team.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,8 @@
             "null"
           ],
           "default": null,
-          "description": "Specifies the folder path to @angular/language-service."
+          "description": "Specifies the folder path to @angular/language-service.",
+          "scope": "machine"
         },
         "angular.log": {
           "type": "string",
--- a/server/src/version_provider.ts
+++ b/server/src/version_provider.ts
@@ -6,6 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import * as fs from 'fs';
+
 const MIN_TS_VERSION = '3.9';
 const MIN_NG_VERSION = '10.0';
 
@@ -25,7 +27,9 @@ function resolve(packageName: string, location: string, rootPackage?: string): N
     const packageJsonPath = require.resolve(`${rootPackage}/package.json`, {
       paths: [location],
     });
-    const packageJson = require(packageJsonPath);
+    // Do not use require() to read JSON files since it's a potential security
+    // vulnerability.
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
     const resolvedPath = require.resolve(packageName, {
       paths: [location],
     });


ddworken:
:wave: Thanks for the tag @kyliau! 

Taking a look at this, I don't think it is quite a complete patch for the bug. While it does fix the POC that I sent, it is still possible to get code execution since if the version is valid, the extension loads the locally installed language server. Ie, when I set up a project with a local `angular.ngdk`, I see the log line:

```
[Info  - 3:28:19 PM] Using @angular/language-service v10.0.7 from /home/david/code/vscode-angular-poc/angular_ndk/node_modules/@angular/language-service/bundles/language-service.js
```

This means if an attacker simply backdoors that file, they can still get code execution through this extension. 

In terms of what a complete fix for this looks like, I think there are two options:

1. Change the `angular.ngdk` setting so it can only be specified in user settings and not in workplace settings. This would prevent a malicious project from changing this setting. This can be done by setting the preferences `scope` to `machine` rather than `resource`. See [these docs](https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema:~:text=A%20configuration%20setting%20can%20have%20one%20of%20the%20following%20possible%20scopes%3A). 

2. Prompt users before running with a workplace configured `angular.ngdk` setting. See [this patch](https://github.com/microsoft/vscode-eslint/commit/b4124f9383121c5f293a1068288e99f77019229d) for an example.

kyliau:
`angular.ngdk` defaults to `window` setting, which according to the docs
> Windows (instance) specific settings which can be configured in user, workspace, or remote settings.

From a security standpoint, what makes user setting more secure than worksplace setting?
Is it because any project in a workspace is allowed to modify the workspace setting?

Does this mean that the native typescript extension in vscode is also susceptible to the same vulnerability?
(they specifically load `tsserver.js`, but the same file could be compromised the same way `language-service.js` is.

https://github.com/microsoft/vscode/blob/eb7a565926ce0c62b18242e8cbb3c3a692db9cb3/extensions/typescript-language-features/package.json#L131-L139

ddworken:
Yeah, the concern is that any project can modify the workspace setting unbeknownst to the user (who might simply want to view a project without running anything). The TypeScript extension avoids this vulnerability by requiring a user to [confirm the setting choice](https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript:~:text=The%20typescript.tsdk%20workspace%20setting%20only%20tells,command%20and%20select%20the%20workspace%20version.) which is similar to my second option described above.

kyliau:
Thanks for the clarification. I've changed the scope of `ngdk` from `window` (default value) to `machine` in a followup commit.
The latter only allows the config value to be modified in user or remote settings.

ddworken:
Looks good to me! :+1:

Issue #909, opened 2020-09-04
Angular Language Server doesn't declare support for workspaceFolders


mickaelistria:
**Describe the bug**

Angular Language Server doesn't declare support for workspaceFolders capability. As a result, IDEs that have used it to determine the multiplicity of the LS (eg Wild Web Developer) now start multiple LS instance as there is nothing to hint that a single instance is enough.

**To Reproduce**

With Wild Web Developer, open multiple .ts files from multiple projects.
See that as many `node .../angular-language-server...` processes are started.
Enabling the langauge server log, the server capabilities only say
```
{"jsonrpc":"2.0","id":"1","result":{"capabilities":{"textDocumentSync":2,"completionProvider":{"resolveProvider":false,"triggerCharacters":["<",".","*","[","(","$","|"]},"definitionProvider":true,"hoverProvider":true}}}
```

kyliau:
Closed by #910

PR #910, opened 2020-09-04
Issue #909 - Explicit support for workspaceFolders

mickaelistria:
As per https://microsoft.github.io/language-server-protocol/specification#workspace_workspaceFolders
, it's recommended that Language Servers that can work with multi-root
workspaces declare support for workspaceFolders.

For the angular language server, it's able to work with any file,
independently of the workspace root; so it's naturally able to work with
multiple roots.
Declaring support for workspaceFolders allows clients that use the
workspaceFolder heuristic as a way to decide of server instance
multiplicity (should it be a single instance or 1 instance per root) to
decide of starting only 1 instance of the LS. Without such information
from the LS, the client integration has to audit the LS, and force some
strategy to decide of the multiplicity, often in an hardcoded way; at
the risk of being incorrect in the future if server changes its own
strategy.
So declaring workspaceFolders allow to automate a better integration in
clients.

Signed-off-by: Mickael Istria <mistria@redhat.com>

### DIFF
--- a/integration/lsp/smoke_spec.ts
+++ b/integration/lsp/smoke_spec.ts
@@ -108,7 +108,8 @@ describe('Angular Language Service', () => {
           'completionProvider':
               {'resolveProvider': false, 'triggerCharacters': ['<', '.', '*', '[', '(', '$', '|']},
           'definitionProvider': true,
-          'hoverProvider': true
+          'hoverProvider': true,
+          'workspace': {'workspaceFolders': {'supported': true}}
         }
       }
     });
@@ -136,7 +137,8 @@ describe('Angular Language Service', () => {
           'completionProvider':
               {'resolveProvider': false, 'triggerCharacters': ['<', '.', '*', '[', '(', '$', '|']},
           'definitionProvider': true,
-          'hoverProvider': true
+          'hoverProvider': true,
+          'workspace': {'workspaceFolders': {'supported': true}}
         }
       }
     });
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -240,6 +240,9 @@ export class Session {
         },
         definitionProvider: true,
         hoverProvider: true,
+        workspace: {
+          workspaceFolders: {supported: true},
+        },
       },
     };
   }


ayazhafiz (commented):
You will need to update https://github.com/angular/vscode-ng-language-service/blob/master/integration/lsp/smoke_spec.ts

integration/lsp/smoke_spec.ts
```diff
@@ -108,7 +108,8 @@ describe('Angular Language Service', () => {
           'completionProvider':
               {'resolveProvider': false, 'triggerCharacters': ['<', '.', '*', '[', '(', '$', '|']},
           'definitionProvider': true,
-          'hoverProvider': true
+          'hoverProvider': true,
+          'workspaceFolders': {'supported': true}
```
ayazhafiz, line 112: This needs to be wrapped in `workspace`:
```suggestion
          'workspace': {
              'workspaceFolders': {'supported': true}
          }
```

ayazhafiz (changes requested):

mickaelistria:
This should now be good.

server/package-lock.json
```diff
@@ -0,0 +1,45 @@
+{
```
ayazhafiz, line 1: Can this file be removed? There is already a `yarn.lock`, and I think it is better to stick with using yarn in this project than both yarn and npm

ayazhafiz (commented):
r+ me after comment

server/package-lock.json
mickaelistria: Sure, I'm removing it. It was actually created/added by mistake.

ayazhafiz (approved):

mickaelistria:
Is there anything missing before this can be merged (and then released)?

kyliau:
@mickaelistria Could you please provide a little more context on what the "workspaceFolders" config does?
It'd be great if you could update the commit message with more info, thank you.

kyliau:
Specifically, what's the use case for this capability and what does it enable?

mickaelistria:
@kyliau I've added some explanation in the commit message.

kyliau:
@mickaelistria Thank you very much for the detailed explanation, it's very helpful! LGTM.

kyliau (approved):

PR #930, opened 2020-09-29
feat: Add typeDefinitionProvider to capabilities

atscott:
See individual commits.

Commit 1: 
Refactor, create some helper functions for common code.

Commit 2: 
The Angular Language Service for Ivy will have a distinction
between "go to definition" and "go to type definition". We
need to add `typeDefinitionProvider` to the capabilities and call
`getTypeDefinitionAtPosition` when type definitions are requested.

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -340,29 +340,26 @@ export class Session {
     }
   }
 
-  private onDefinition(params: lsp.TextDocumentPositionParams) {
-    const {position, textDocument} = params;
-    const filePath = uriToFilePath(textDocument.uri);
-    const scriptInfo = this.projectService.getScriptInfo(filePath);
-    if (!scriptInfo) {
-      this.connection.console.log(`Script info not found for ${filePath}`);
+  private onDefinition(params: lsp.TextDocumentPositionParams): lsp.LocationLink[]|undefined {
+    const lsInfo = this.getLSAndScriptInfo(params.textDocument);
+    if (lsInfo === undefined) {
       return;
     }
-
-    const {fileName} = scriptInfo;
-    const langSvc = this.getDefaultLanguageService(scriptInfo);
-    if (!langSvc) {
-      return;
-    }
-
-    const offset = lspPositionToTsPosition(scriptInfo, position);
-    const definition = langSvc.getDefinitionAndBoundSpan(fileName, offset);
+    const {languageService, scriptInfo} = lsInfo;
+    const offset = lspPositionToTsPosition(scriptInfo, params.position);
+    const definition = languageService.getDefinitionAndBoundSpan(scriptInfo.fileName, offset);
     if (!definition || !definition.definitions) {
       return;
     }
     const originSelectionRange = tsTextSpanToLspRange(scriptInfo, definition.textSpan);
+    return this.tsDefinitionsToLspLocationLinks(definition.definitions, originSelectionRange);
+  }
+
+  private tsDefinitionsToLspLocationLinks(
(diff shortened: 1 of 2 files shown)


server/src/session.ts
```diff
@@ -386,22 +381,34 @@ export class Session {
     return results;
   }
 
-  private onHover(params: lsp.TextDocumentPositionParams) {
+  private getLSPositionInfo(params: lsp.TextDocumentPositionParams):
```
ayazhafiz, line 399: I would prefer either just returning a `scriptInfo` from this function and having the caller create a language service and getting the offset, or rename the function to something like `getLSAndScriptPosition`.

server/src/session.ts
```diff
@@ -340,29 +340,25 @@ export class Session {
...
     if (!definition || !definition.definitions) {
       return;
     }
     const originSelectionRange = tsTextSpanToLspRange(scriptInfo, definition.textSpan);
+    const results: lsp.LocationLink[] = this.definitionsToLSPLocation(definition.definitions);
+    return results.map(r => ({...r, originSelectionRange}));
+  }
+
+  private definitionsToLSPLocation(definitions: readonly ts.DefinitionInfo[]):
+      Array<{targetUri: string, targetRange: lsp.Range, targetSelectionRange: lsp.Range}> {
```
ayazhafiz, line 374: Could we pull the `Array` type parameter out as an interface and add comments about what each are? I realize this is outside the scope of the PR so feel free to disregard, just suggesting because you are already here :)

ayazhafiz (approved):
LGTM besides a couple comments.

Note: calling `getTypeDefinition` from an LSP host currently fails (non-fatally):

```
[Error - 2:41:03 PM] Request textDocument/typeDefinition failed.
  Message: Request textDocument/typeDefinition failed with message: Could not find source file: '/Users/hafiz/Code/Angular/example/src/app/app.component.html'.
  Code: -32603 
```

I guess this makes sense bc the Angular LS does not actually implement getting a type definition yet.

server/src/session.ts
```diff
@@ -340,29 +342,38 @@ export class Session {
...
-    const originSelectionRange = tsTextSpanToLspRange(scriptInfo, definition.textSpan);
+    const {languageService, scriptInfo, offset} = lsInfo;
+    const definitions = languageService.getTypeDefinitionAtPosition(scriptInfo.fileName, offset);
+    if (!definitions) {
+      return;
+    }
+    return this.definitionsToLSPLocation(definitions);
+  }
+
+  private definitionsToLSPLocation(definitions: readonly ts.DefinitionInfo[]):
```
kyliau, line 373: 1. Why not just return `lsp.LocationLink[]`?
2. Please consider renaming this to `tsDefinitionsToLspLocationLinks`

server/src/session.ts
```diff
@@ -239,6 +240,7 @@ export class Session {
           triggerCharacters: ['<', '.', '*', '[', '(', '$', '|']
         },
         definitionProvider: true,
+        typeDefinitionProvider: true,
```
kyliau, line 243: This means we will have to also implement `getTypeDefinitionAtPosition` for View Engine.
Please make the change in `@angular/language-service` first before merging this PR.

server/src/session.ts
atscott: 1. I wanted to be more explicit about what was returned. Updated to accept optional `originSelectionRange`
2. done

server/src/session.ts
atscott: Done.

kyliau (approved):

PR #951, opened 2020-11-02
refactor: move notification types to common package

kyliau:
Currently, the notification types in `protocol.ts` are duplicated in both
`client` and `server`.

This commit consolidates them into `common` and create a project reference.
This is done in preparation for the ngcc work, which will add more notification
types.

With this change, `client` and `server` now have to depend on the `.d.ts`
output of `common`.
Although this is the right way to organize the code, it introduces two
unfortunate side effects:

1. `client` and `server` now have to import from `common/out`, since that's
   the output directory where the declaration files are emitted.
2. users who clone this repo will have to run `yarn compile` to build the
   repo before they can navigate the project in their editor.

(1) does not look "nice", but it offers a straight-forward way without too
drastic changes to the build workflow.

While (2) is not ideal, it is the correct way of using project references
according to the TS [documentation](https://www.typescriptlang.org/docs/handbook/project-references.html#caveats-for-project-references).

> Because dependent projects make use of .d.ts files that are built from their
> dependencies, you’ll either have to check in certain build outputs or build
> a project after cloning it before you can navigate the project in an editor
> without seeing spurious errors.

TypeScript does plan to improve the ergonomics of project references in the
future.

> We’re working on a behind-the-scenes .d.ts generation process that should be
> able to mitigate this, but for now we recommend informing developers that
> they should build after cloning.


kyliau:
@ayazhafiz @atscott @alxhub Could you please give me some feedback on this change?

It looks simple, but getting the code organization right is surprisingly challenging, and this is the most straight-forward way I could think of.

The alternative is to create a true private package for the common code, and import the package in `client` and `server` via module specifier, like we do in `@angular/angular`.
However, doing so will make this repo a true monorepo, and it comes with its own set of problems:
1. We'll have to duplicate dependencies in root package.json and in individual package's package.json, AND keep them in sync. We could use yarn workspaces to simplify the setup, but it is definitely non-trivial.
2. We'll have to drastically alter the dev workflow so that "F5" continues to build the code and its transitive dependencies before launching the development instance. We'll have to also make sure node module resolution works as expected.
3. We'll have to modify the build setup so that rollup continues to work correctly.

So ... I guess you can tell I am in favor of the more straight-forward approach, at the expense of the two side effects mentioned in the PR description.

ayazhafiz (commented):
I think this is good, I am in favor of keeping it as simple as possible. It sounds like the only real impact here is on TS language services, but I don't think this is a big deal if the common module is just notification IDs.

Could you add a note to the dev documentation describing that common needs to be compiled separately for language services?

kyliau:
> I think this is good, I am in favor of keeping it as simple as possible. It sounds like the only real impact here is on TS language services, but I don't think this is a big deal if the common module is just notification IDs.
> 
> Could you add a note to the dev documentation describing that common needs to be compiled separately for language services?

Updated `DEVELOPER.md`

ayazhafiz (approved):
:+1: this looks great to me, Keen. Excited to see the ngcc addition.

atscott (approved):
@kyliau Thanks for the really great description of the tradeoffs here. I think this is the right decision. I feel like the complexity of the project organization should mirror the complexity of the code and dependencies. Because it's such a small piece and it's only imported in one place in each of the client and server, it makes sense to me to keep it as simple as possible. It doesn't seem worth the investment in the alternative you mentioned (private package) at this time.

PR #955, opened 2020-11-04
test: add angularCompilerOptions to test project

kyliau:
This commit adds `angularCompilerOptions` for Ivy in the test project.

### DIFF
--- a/integration/project/tsconfig.json
+++ b/integration/project/tsconfig.json
@@ -6,5 +6,9 @@
     "typeRoots": [
       "node_modules/@types"
     ]
+  },
+  "angularCompilerOptions": {
+    "strictTemplates": true,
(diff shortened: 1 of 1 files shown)


ayazhafiz:
What is the purpose of this in the extension tests?

kyliau:
I use this as the test project for the extension in dev mode, and thought it's worthwhile to make the project compatible with Ivy LS as well.
It's the same configuration used in 
https://github.com/angular/angular/blob/master/packages/language-service/test/project/tsconfig.json

As a follow-up, I could write some new integration tests for Ivy

atscott (approved):

ayazhafiz (approved):
Got it, thank you for the explanation.

PR #958, opened 2020-11-05
test: refactor integration test for LSP

kyliau:
It is not always straight-forward to write unit tests for the code in
this repo because of the nature of the language server.

However, since the language server uses JSON-RPC, it makes sense to test
the server along this boundary through stateful integration test.

This PR makes writing integration tests a lot easier by refactoring the
test client and disabling Jasmine randomization. As a result, we can now

1. Use strongly typed requests and responses.
2. Have much better abstraction of the test client.
3. Easily write more tests!

This refactoring is done in preparation for the upcoming ngcc feature.


ayazhafiz (commented):
I'm not sure making the tests stateful is a good idea, since this means adding later tests with more complex interactions could be more difficult to write and/or debug (would have to change the correct files previously modified in tests).

To decrease the verbosity in writing tests, I think we could pull out some common things like calling `initialize` and `shutdown` to `createConnection` or similar utility functions. Since the  server options should always be the same, the response of initialize could be validated there. Similarly utility functions for `didOpen`, `didChange`, etc. could be written to avoid having to write out the entire JSON request. What do you think?

atscott:
> I'm not sure making the tests stateful is a good idea, since this means adding later tests with more complex interactions could be more difficult to write and/or debug (would have to change the correct files previously modified in tests).
> 
> To decrease the verbosity in writing tests, I think we could pull out some common things like calling `initialize` and `shutdown` to `createConnection` or similar utility functions. Since the server options should always be the same, the response of initialize could be validated there. Similarly utility functions for `didOpen`, `didChange`, etc. could be written to avoid having to write out the entire JSON request. What do you think?

Yep, I agree with Ayaz here. I was in the middle of adding my own comment saying just about the same thing.

kyliau:
My initial intention of turning off randomization is to improve test performance.
With randomization off, we could initialize the server once, and perform a series of operations much like an IDE would.
With randomization on, we get better test hermeticity, at the expense of long server startup for each spec.

Since the number of tests are very small right now, I agree with both of you that we should prefer hermeticity over performance. I've reenabled the randomization.

integration/lsp/jasmine.json
```diff
@@ -0,0 +1,6 @@
+{
```
kyliau, line 1: I'm keeping this config file around to make it easier to turn off randomization when debugging.

kyliau:
@atscott and I talked about creating a server instance that persists across tests, but reset the state after every test.
While there's no LSP method for reloading the project, we could make the test client close all the open files after every test, and never save them. @ayazhafiz what do you think?

I think we should prefer a straightforward solution for now (either relaunch the server before every test or close the open files), and revise this strategy once we have more meaningful tests.

ayazhafiz:
> While there's no LSP method for reloading the project, we could make the test client close all the open files after every test, and never save them. @ayazhafiz what do you think?

I think this could work, but there is the issue of what the project service on the server side will do (since it's still the same). Namely, does the project service destroy the script info for a file when we close it? If we delete and then recreate a tsconfig.json, does the project service close reuse the last project? Maybe it doesn't matter if all of this is reused, but it needs some investigation I think.
Also, I could see the possibility that in the (maybe near) future we have a variety of client options we might want to initialize with (maybe something similar to [this tsserver extension's options](this.projectService.closeClientFile)), in which case we would need to reinitialize the server often.

> I think we should prefer a straightforward solution for now (either relaunch the server before every test or close the open files), and revise this strategy once we have more meaningful tests.

I agree, I am in favor of keeping it as simple + flexible as possible by just relaunching the server. I think it is too early to try to think about optimizations, and my guess would be that bottlenecks we see are with creating a new project, not spawning a new server, but as I mentioned before I think we might have some issues if we try to reuse the same project.

kyliau:
> does the project service destroy the script info for a file when we close it?

It does not destroy it immediately, but the next time `open` is called, the text we pass to the server will overwrite the existing content of the script info.

> Also, I could see the possibility that in the (maybe near) future we have a variety of client options we might want to initialize with (maybe something similar to [this tsserver extension's options](this.projectService.closeClientFile)), in which case we would need to reinitialize the server often.

Yeah, I agree. Keeping each test hermetic would make it easier to test different scenarios.

> I agree, I am in favor of keeping it as simple + flexible as possible by just relaunching the server.

OK, as this PR currently is, each test initializes its own server instance, so I think it's good to go. PTAL.

integration/lsp/test_utils.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export function initializeServer(client: MessageConnection): Promise<lsp.InitializeResult> {
+  return client.sendRequest(lsp.InitializeRequest.type, {
+    /**
+     * The process Id of the parent process that started
+     * the server. Is null if the process has not been started by another
+     * process. If the parent process is not alive then the server should
+     * exit (see exit notification) its process.
+     */
```
ayazhafiz, line 56: Is this comment copied from the specification or somewhere else? I think we can simplify it to just something like
```suggestion
    /**
     * The process id of the parent process that started
     * the server. Is always the current process.
     */
```
for our use.

ayazhafiz (approved):

integration/lsp/viewengine_spec.ts
```diff
@@ -0,0 +1,111 @@
...
+        },
+      ],
+    });
+    const response = await client.sendRequest(lsp.CompletionRequest.type, {
+      textDocument: {
+        uri: `file://${APP_COMPONENT}`,
+      },
+      position: {line: 4, character: 30},
+    });
+    expect(response).toContain({
```
atscott, line 74: I think these expectations are fine for a first pass, but let's keep an eye on these for when there are more tests. These expectations aren't easy to understand and make it hard to derive meaning from them. Here are a couple things that I can think that would improve them with a `humanize` that does:
* Convert text ranges to actual text
* convert `kind` to a meaningful string
* I don't know what the meaning is for `label`, `sortText`, or `newText`. Maybe we don't need to assert all of those
*  The meaning of `detail` is related to where the text changed; we might want to improve the ways in which we construct the `sendNotification` object to also be more human readable so one can easily see "Oh, I'm adding a '.' char after XYZ text (I don't want to open the file and go to the line/char to figure that out), and I would expect the result to be X based on that."

atscott (approved):

integration/lsp/viewengine_spec.ts
kyliau: Yeah, we should definitely revisit how these integration tests are done.
Initially, what I had in mind was more like golden tests, where we just dump the entire response to JSON.
In fact, that's what I did for the integration test in `@angular/angular`: https://github.com/angular/angular/tree/master/integration/language_service_plugin/goldens

Although that test looks very similar to this one, they're different in one key way:
The one in `@angular/angular` is for tsserver **plugin** whereas this is for the LSP server.

We should probably unify if not consolidate these tests.

integration/lsp/viewengine_spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    });
+    client.listen();
+  });
+
+  afterEach(() => {
+    client.dispose();
+  })
+
+  it('should handle initialization at startup', async () => {
+    const response = await initializeServer(client);
```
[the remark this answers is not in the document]
kyliau: Good idea, thanks!

PR #959, opened 2020-11-06
Properly inject Angular template grammar in inline templates

### LINKED ISSUE #768: Syntax coloring of inline templates
This syntax coloring in my inline template is starting to make my eyes bleed:

Screen Shot 2020-05-07 at 9 37 02 PM

Inline templates are all we have because it greatly reduces file clutter. But this is a nasty price to pay. What's involved in improving this?

ayazhafiz:
It seems that to get inline templates to be syntax highlighted with the
angular template grammar correctly, we not only need to include the
`template.ng` grammar pattern when matching an inline template (which
previously worked only for interpolations but would not seem to
correctly recognize template bindings); we also need to tell VSCode that
it should inject the template grammar in TypeScript files.

I cannot find a way to test this (since it differs from the TextMate
grammar tests, as this change instructs VSCode specifically how to add
the grammars), but screenshots attached to the PR for this commit
demonstrate before- and after-this-change syntax highlight on an inline
template. Furthermore non-template TypeScript should not be matched by
the template grammar, since the template grammar can only be injected
when the base grammar is matched to be HTML (which only happens when we
match an inline template).

I couldn't figure this out a couple months back, but looked back today
and got it almost immediately :) I guess some time away is a good thing.

Closes #768

Screen Shot 2020-11-05 at 8 58 49 PM
Screen Shot 2020-11-05 at 8 58 06 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,8 @@
         "path": "./syntaxes/template.json",
         "scopeName": "template.ng",
         "injectTo": [
-          "text.html.derivative"
+          "text.html.derivative",
+          "source.ts"
         ],
         "embeddedLanguages": {
           "text.html": "html",


ayazhafiz:
(atscott, alxhub not sure that you are interested, but adding you just in case you are)

kyliau (approved):

PR #960, opened 2020-11-06
Properly parse nested ternary expressions

### LINKED ISSUE #957: in the html page,some code's color is wrong
color error
in line 168, below the line,the code's color becomes green, if i delete the 'click',the color will become right.
but above , there also are 'click's,just begin from this line and this 'click', it looks like strange .

ayazhafiz:
The previous implementation of the template grammar for ternary
expressions would not match nested ternary expressions. In this commit
we update the matcher to use the TypeScript TM grammar's implementation
to do this correctly. Screenshots attached to the PR for this commit
demonstrate the change in a "real" template.

Closes #957
Screen Shot 2020-11-05 at 9 40 32 PM
Screen Shot 2020-11-05 at 9 39 59 PM

### DIFF
--- a/syntaxes/src/expression.ts
+++ b/syntaxes/src/expression.ts
@@ -593,8 +593,12 @@ export const Expression: GrammarDefinition = {
       match: /\\(x\h{2}|[0-2][0-7]{0,2}|3[0-6][0-7]?|37[0-7]?|[4-7][0-7]?|.|$)/,
     },
 
+    // Derived from
+    // https://github.com/sheetalkamat/TypeScript-TmLanguage-VsCode/blob/04219109f3e2090f787c0fa8743e73a4bd1ad876/syntaxes/TypeScript.tmLanguage.json#L2830-L2848
     ternaryExpression: {
-      begin: /(?!\?\.\s*[^[:digit:]])(\?)(?!\?)/,
+      // 1. negative lookahead (?!\?\.\s*[^[:digit:]]) ensures safe navigation does not match
+      // 2. match the ternary question mark with (\?)
+      begin: /(?!\?\.\s*[^[:digit:]])(\?)/,
       beginCaptures: {
         1: {
           name: 'keyword.operator.ternary.ts',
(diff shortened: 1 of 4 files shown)


syntaxes/src/expression.ts
```diff
@@ -594,28 +594,15 @@ export const Expression: GrammarDefinition = {
...
-        },
-        {
-          include: '#ngExpression',
-        },
-      ],
-    },
-
-    ternaryOperator: {
-      begin: /(\?)(?!\.)/,
+      begin: /(?!\?\.\s*[^[:digit:]])(\?)(?!\?)/,
```
atscott, line 597: Since this file is a TS file, could you add a comment explaining the regex groups:
1. negative lookahead `(?!\?\.\s*[^[:digit:]])` ensures save navigation does not match
2. match the ternary question mark with `(\?)`
3. negative lookahead after the first question mark to ensure we don't match nullish coalesce `(?!\?)`

If you don't want to add those here, could you add comments like this to the ternary test cases?

syntaxes/test/data/expression.html
```diff
@@ -70,6 +70,7 @@
 {{ condition ? ['123'] : { test: 'a' } }}
 {{ condition ? 'test' : "test" }}
 {{ condition ? [function(), variable] : {} }}
+{{ condition ? condition2 ? condition3 ? 1 : 2 : 3 : 4 }}
```
atscott, line 73: Looks like you need a test for the nullish coalesce case (`??`). Since this test doesn't exist but the case is covered in the regex, I'm curious if you grabbed the regex from another source that we can link to if it's something "official" like the TypeScript repo. You mentioned "TypeScript TM grammar's implementation" in the PR comment - is that something we can link to?

atscott (approved):

syntaxes/test/data/expression.html
ayazhafiz: Good catch, we should remove that in the first place since the template grammar doesn't have the `??` syntax.

syntaxes/src/expression.ts
ayazhafiz: Done

PR #962, opened 2020-11-07
fix: log all console output to file

kyliau:
We currently ask users to submit two logs when opening a new issue:

1. Logs from file, and
2. Logs from output channel (see screenshot below).

In order to simplify the workflow, we should log all console output to file
as well. This has two advantages:

1. We can piece together a more coherent picture by just looking at the log
   file.
2. Users no longer have to submit logs from output channel.

Additional context: when debugging the integration test, I personally found it difficult to get a full understanding by just looking at the log file.

Screen Shot 2020-11-06 at 8 16 59 PM


ayazhafiz (approved):
After the next release is cut we should update the issue template to ask only for the log file.

Issue #922, opened 2020-09-21
Error in HTML not reported on file open


mickaelistria:
Using @angular/language-server 0.1001.0 and @angular/language-service 10.1.1, I see a regression compared to @angular/language-server 0.1000.4.
To reproduce, with Eclipse WIld Web Developer
1. open and provision (`npm install`) a simple project like https://github.com/eclipse/wildwebdeveloper/tree/master/org.eclipse.wildwebdeveloper.tests/testProjects/angular-app
2. Open appComponentWithHtml.ts , let tsserver and angular-ls load (ie show outline or provide some Angular specific completion)
3. Then open appConponentWithHtml.html
  EXPECTED: Error marker shows on incorrect template
  GOT: No error marker. Note that if I then make any change to the document, the expected error marker shows up.

With previous versions of @angular/language-server, this was working as expected, the marker/diagnostic was showing up on file open.

mickaelistria:
Actually, the issue doesn't come from migration to newer @angular/language-serv*, but from simultaneous migration from typescript 3.9.7 to typescript 4.0.2. If using typescript 3.9.7, all @angular/language-serv* versions mentioned earlier works perfectly; if using typescript 4.0.2, all @angular/language-serv* versions mentioned earlier fail at showing diagnostic on .html file opening (but do show it after edit).

PR #966, opened 2020-11-09
fix: LS not showing existing diagnostics on file open

kyliau:
This commit fixes a bug whereby existing diagnostics in an external template
are not shown when the file is opened the first time. Diagnostics only show
up when subsequent edits are made to the file.
This is a regression after upgrading to TS 4.0

Fix https://github.com/angular/vscode-ng-language-service/issues/922

### DIFF
--- a/integration/lsp/viewengine_spec.ts
+++ b/integration/lsp/viewengine_spec.ts
@@ -86,6 +86,25 @@ describe('Angular language server', () => {
       },
     });
   });
+
+  it('should show existing diagnostics on external template', async () => {
+    client.sendNotification(lsp.DidOpenTextDocumentNotification.type, {
+      textDocument: {
+        uri: `file://${FOO_TEMPLATE}`,
+        languageId: 'typescript',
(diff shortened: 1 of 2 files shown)


ayazhafiz (approved):
Why did this change? Is it to do with `externalFiles`?

kyliau:
@ayazhafiz https://github.com/microsoft/TypeScript/pull/37239 changed the implementation of `openClientFile` such that it no longer returns the `configFileName` when the opening of a file causes a new project to be created. 
We actually saw this behavior earlier in https://github.com/angular/vscode-ng-language-service/pull/935. While the change is harmless on its own because we already handle subsequent project discovery with `ProjectService.getDefaultProjectForScriptInfo()`, I overlooked the fact that we also trigger diagnostics on file open. We should not return early when `configFileName` is absent. Instead we should work around it by looking at the `containingProjects` of the `scriptInfo` created by the call to `openClientFiles()`.

atscott (approved):

PR #967, opened 2020-11-10
feat: server-side implementation for ngcc feature

kyliau:
Ivy compiler requires Ivy-specific properties generated by ngcc to be
available in the TS declaration files. As a result, the language service
will have to run ngcc on project load.

Since ngcc modifies node_modules, it should be run on the client side.

This commit implements the server-side code required to support ngcc.
The idea is as follows:

1. After a project has finished loading, the server sends a notification
   to the client to run ngcc. Meanwhile, the server keeps the language
   service disabled.
2. Client runs ngcc. Once complete, client sends another notification to the
   server. Server re-enables language service.

### DIFF
--- a/integration/lsp/ivy_spec.ts
+++ b/integration/lsp/ivy_spec.ts
@@ -1,0 +1,76 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {MessageConnection} from 'vscode-jsonrpc';
+
+import {APP_COMPONENT, createConnection, initializeServer, openTextDocument} from './test_utils';
+
+describe('Angular Ivy language server', () => {
+  jasmine.DEFAULT_TIMEOUT_INTERVAL = 10000; /* 10 seconds */
+
+  let client: MessageConnection;
+
+  beforeEach(() => {
+    client = createConnection({
+      ivy: true,
+    });
+    client.listen();
+  });
+
+  afterEach(() => {
+    client.dispose();
+  });
+
+  it('should send ngcc notification after a project has finished loading', async () => {
+    await initializeServer(client);
+    openTextDocument(client, APP_COMPONENT);
+    const configFilePath = await onRunNgccNotification(client);
+    expect(configFilePath.endsWith('integration/project/tsconfig.json')).toBeTrue();
+  });
+
+  it('should disable language service until ngcc has completed', async () => {
+    await initializeServer(client);
+    openTextDocument(client, APP_COMPONENT);
+    const languageServiceEnabled = await onLanguageServiceStateNotification(client);
+    expect(languageServiceEnabled).toBeFalse();
+  });
+
+  it('should re-enable language service once ngcc has completed', async () => {
+    await initializeServer(client);
+    openTextDocument(client, APP_COMPONENT);
+    const configFilePath = await onRunNgccNotification(client);
+    client.sendNotification('angular/ngccComplete', {
+      success: true,
+      configFilePath,
+    });
+    const languageServiceEnabled = await onLanguageServiceStateNotification(client);
+    expect(languageServiceEnabled).toBeTrue();
+  });
+});
+
+function onRunNgccNotification(client: MessageConnection): Promise<string> {
+  return new Promise(resolve => {
+    // TODO(kyliau): Figure out how to import the notification type from
+    // common/out/notifications.d.ts. Currently we cannot do this because the
+    // TS files and JS outputs are in different trees. As a result, node module
+    // resolution works for the former but not the latter since their relative
+    // import paths are different.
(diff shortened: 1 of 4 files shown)


integration/lsp/ivy_spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  });
+});
+
+function onRunNgccNotification(client: MessageConnection): Promise<string> {
+  return new Promise(resolve => {
+    // TODO(kyliau): Figure out how to import the notification type from
+    // common/out/notifications.d.ts. Currently we cannot do this because the
+    // TS files and JS outputs are in different trees. As a result, node module
+    // resolution works for the former but not the latter since their relative
+    // import paths are different.
```
ayazhafiz, line 62: Side note, no action required at this time: I think we need to get rid of the `common/out` pattern entirely and instead provide a separate package. Will discuss this more in our sync.

integration/lsp/ivy_spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    client.onNotification('angular/runNgcc', (params: {configFilePath: string}) => {
```
ayazhafiz, line 63: Can this and `angular/projectLanguageService` be pulled out as constants? I believe the language client provides a notification type where we could name this and list the expected parameters?

server/src/session.ts
```diff
@@ -102,6 +105,24 @@ export class Session {
...
     conn.onHover(p => this.onHover(p));
     conn.onCompletion(p => this.onCompletion(p));
+    conn.onNotification(notification.NgccComplete, p => this.handleNgccNotification(p));
+  }
+
+  private handleNgccNotification(params: notification.NgccCompleteParams) {
+    const {configFilePath} = params;
+    if (!params.success) {
+      this.error(
+          `Failed to run ngcc for ${configFilePath} because ${params.error}. ` +
```
ayazhafiz, line 115:
```suggestion
          `Failed to run ngcc for ${configFilePath}:\n` +
          ${params.error}\n` +
```

server/src/session.ts
```diff
@@ -102,6 +105,24 @@ export class Session {
...
+          `Language service will remain disabled.`);
+      return;
+    }
+    const project = this.projectService.findProject(configFilePath);
+    if (!project) {
+      this.error(`Failed to find project for ${configFilePath} returned by ngcc.`);
```
ayazhafiz, line 121:
```suggestion
      this.error(`Failed to find project for ${configFilePath} returned by ngcc.\nLanguage service will remain disabled.`);
```

server/src/session.ts
```diff
@@ -118,23 +139,23 @@ export class Session {
...
       case ts.server.ProjectsUpdatedInBackgroundEvent:
         // ProjectsUpdatedInBackgroundEvent is sent whenever diagnostics are
         // requested via project.refreshDiagnostics()
         this.triggerDiagnostics(event.data.openFiles);
         break;
+      case ts.server.ProjectLanguageServiceStateEvent:
+        this.connection.sendNotification(notification.ProjectLanguageService, {
+          projectName: event.data.project.getProjectName(),
+          languageServiceEnabled: event.data.languageServiceEnabled,
+        });
```
ayazhafiz, line 158: The client doesn't actually use this right? This is just for the tests? (Making sure my mental model is correct)

server/src/session.ts
```diff
@@ -517,42 +536,51 @@ export class Session {
...
-    // The language service should be enabled at this point.
-    this.info(`Enabling language service for ${projectName}.`);
+    if (this.ivy && project instanceof ts.server.ConfiguredProject) {
+      // Keep language service disabled until ngcc is completed.
+      project.disableLanguageService();
+      this.connection.sendNotification(notification.RunNgcc, {
+        configFilePath: project.getConfigFilePath(),
+      });
+    } else {
+      // The language service should be enabled at this point.
```
ayazhafiz, line 551:
```suggestion
      // Immediately enable Legacy/ViewEngine language service
```

atscott (approved):

integration/lsp/ivy_spec.ts
kyliau: yeah, we might have to go the monorepo route to properly use `common` as a package. It'll require some major changes to the build.

integration/lsp/ivy_spec.ts
kyliau: I think your comment here relates to the TODO?

server/src/session.ts
kyliau: Yeah, it's just for tests in our setup. Although our vscode client doesn't do anything for this notification, I figured it's still useful to expose this for other clients. Notifications are cheap and non-blocking, so it's not too much overhead.

integration/lsp/ivy_spec.ts
ayazhafiz: oops I missed that. yes, thanks.

ayazhafiz (approved):

PR #972, opened 2020-11-12
feat: client-side implementation for ngcc feature

kyliau:
This commit implements the client-side work to support the ngcc feature.

The idea is as follows:

1. Client receives notification from server
2. Client runs ngcc
3. Client sends notification to server once ngcc is done

Watch screen recording [here](https://drive.google.com/file/d/1U1RpRD7SOKfx3E0uOjKh36LyqCkiSTiF/view?usp=sharing)

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -11,9 +11,11 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import * as lsp from 'vscode-languageclient';
 
-import {projectLoadingNotification} from '../../common/out/notifications';
+import * as notification from '../../common/out/notifications';
 
+import {resolveAndRunNgcc} from './command-ngcc';
 import {registerCommands} from './commands';
+import {withProgress} from './progress-reporter';
 
 export function activate(context: vscode.ExtensionContext) {
   // If the extension is launched in debug mode then the debug server options are used
@@ -53,29 +55,50 @@ export function activate(context: vscode.ExtensionContext) {
   // client can be deactivated on extension deactivation
   context.subscriptions.push(...registerCommands(client), client.start());
 
-  client.onDidChangeState((e) => {
-    let task: {resolve: () => void}|undefined;
-    if (e.newState == lsp.State.Running) {
-      client.onNotification(projectLoadingNotification.start, () => {
-        if (task) {
-          task.resolve();
-          task = undefined;
-        }
-        vscode.window.withProgress(
-            {
-              location: vscode.ProgressLocation.Window,
-              title: 'Initializing Angular language features',
-            },
-            () => new Promise((resolve) => {
-              task = {resolve};
-            }));
-      });
+  client.onDidChangeState((e: lsp.StateChangeEvent) => {
+    if (e.newState === lsp.State.Running) {
+      registerNotificationHandlers(client);
+    }
+  });
+}
 
-      client.onNotification(projectLoadingNotification.finish, () => {
-        if (task) {
-          task.resolve();
-          task = undefined;
-        }
+function registerNotificationHandlers(client: lsp.LanguageClient) {
+  client.onNotification(notification.ProjectLoadingStart, () => {
+    vscode.window.withProgress(
+        {
+          location: vscode.ProgressLocation.Window,
+          title: 'Initializing Angular language features',
+        },
+        () => new Promise((resolve) => {
+          client.onNotification(notification.ProjectLoadingFinish, resolve);
+        }),
+    );
+  });
+
+  client.onNotification(notification.RunNgcc, async (params: notification.RunNgccParams) => {
+    const {configFilePath} = params;
+    try {
+      await withProgress(
+          {
+            location: vscode.ProgressLocation.Window,
+            title: `Running ngcc for project ${configFilePath}`,
+            cancellable: false,
+          },
+          (progress: vscode.Progress<string>) => {
+            return resolveAndRunNgcc(configFilePath, progress);
+          },
+      );
+      client.sendNotification(notification.NgccComplete, {
+        configFilePath,
+        success: true,
+      });
+    } catch (e) {
+      vscode.window.showWarningMessage(
+          `Failed to run ngcc. Ivy language service might not function correctly. Please see the log file for more information.`);
+      client.sendNotification(notification.NgccComplete, {
+        configFilePath,
+        success: false,
+        error: e.message,
       });
     }
   });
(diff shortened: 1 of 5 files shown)


client/src/extension.ts
```diff
@@ -53,29 +55,62 @@ export function activate(context: vscode.ExtensionContext) {
...
+    }
+  });
+
+  client.onNotification(notification.RunNgcc, async (params: notification.RunNgccParams) => {
+    const {configFilePath} = params;
+    try {
+      await withProgress(
+          {
+            location: vscode.ProgressLocation.Window,
+            title: `Running ngcc on ${configFilePath}`,
```
ayazhafiz, line 96:
```suggestion
            title: `Running ngcc for project ${configFilePath}`,
```

Also, can we make this path relative to the open WS directory to make UI a little bit clearer?

client/src/extension.ts
```diff
@@ -53,29 +55,62 @@ export function activate(context: vscode.ExtensionContext) {
...
+            return resolveAndRunNgcc(configFilePath, progress);
+          },
+      );
+      client.sendNotification(notification.NgccComplete, {
+        configFilePath,
+        success: true,
+      });
+    } catch (e) {
+      vscode.window.showWarningMessage(
+          `Failed to run ngcc. Ivy language service might not function correctly. Please see log files for more information.`)
```
ayazhafiz, line 109:
```suggestion
          `Failed to run ngcc. Ivy language service might not function correctly. Please see the log file for more information.`)
```

client/src/extension.ts
```diff
@@ -53,29 +55,62 @@ export function activate(context: vscode.ExtensionContext) {
...
+  });
+
+  client.onNotification(notification.ProjectLoadingFinish, () => {
+    if (task) {
+      task.resolve();
+      task = undefined;
+    }
+  });
+
+  client.onNotification(notification.RunNgcc, async (params: notification.RunNgccParams) => {
```
ayazhafiz, line 90: Do we need `task.resolve()` here?

client/src/extension.ts
```diff
@@ -53,29 +55,62 @@ export function activate(context: vscode.ExtensionContext) {
...
-          task.resolve();
-          task = undefined;
-        }
+function registerNotificationHandlers(client: lsp.LanguageClient) {
+  let task: {resolve: () => void}|undefined;
+  client.onNotification(notification.ProjectLoadingStart, () => {
+    if (task) {
+      task.resolve();
+      task = undefined;
+    }
```
ayazhafiz, line 71: nit
```suggestion
      task?.resolve();
      task = undefined;
```

Or even pull out into a function `closeLastTask`

client/src/progress-reporter.ts
```diff
@@ -0,0 +1,50 @@
...
+import * as vscode from 'vscode';
+
+const EMPTY_DISPOSABLE = vscode.Disposable.from();
+
+class ProgressReporter implements vscode.Progress<unknown> {
+  private disposable: vscode.Disposable = EMPTY_DISPOSABLE;
+
+  report(value: unknown) {
+    this.disposable.dispose();  // clear the last message
+    this.disposable = vscode.window.setStatusBarMessage(`$(loading~spin) Angular: ${value}`)
```
ayazhafiz, line 18: Wow, they have their own formatter for the message? (!)

client/src/progress-reporter.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as vscode from 'vscode';
+
+const EMPTY_DISPOSABLE = vscode.Disposable.from();
+
+class ProgressReporter implements vscode.Progress<unknown> {
+  private disposable: vscode.Disposable = EMPTY_DISPOSABLE;
```
ayazhafiz, line 14: I was confused by what a "disposable" was until I looked at https://vscode-docs.readthedocs.io/en/stable/extensions/patterns-and-principles/#disposables. Could we name this `lastMessage` or `disposableMsg` or similar?

ayazhafiz (commented):
Keen, did you check in command-ngcc.ts? Looks like it is not in the commit.

client/src/progress-reporter.ts
kyliau: They do! Here's the reference docs: https://code.visualstudio.com/api/references/icons-in-labels
I should add the link in the comment as well.

client/src/progress-reporter.ts
kyliau: yeah, disposables are prevalent in the vscode codebase. I guess it's the [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) equivalent in Javascript.
I've renamed this `lastMessage`.

client/src/extension.ts
kyliau: No we don't need it here, the `task` is used to clear the notification for `ProjectLoadingStart`.
I realized could consolidate this into the function above, so we don't need `task` anymore.

client/src/progress-reporter.ts
ayazhafiz: I very much miss RAII in languages like JS and Python. But I digress..

kyliau:
> Keen, did you check in command-ngcc.ts? Looks like it is not in the commit.

Oops. I did! This means our test coverage is not good at all for the client.. I'll need to figure out a way to test this.
We have the "e2e" test but it doesn't work on CI.

ayazhafiz (commented):
Nice! Are there tests for this?

ayazhafiz:
> This means our test coverage is not good at all for the client.. I'll need to figure out a way to test this.
> We have the "e2e" test but it doesn't work on CI.

At the very least a build test could be added, but I thought we already did that.

atscott (approved):
👍  I like the actor model-esque design here with message passing. Excited to see this come together!

It would be nice to see tests for this, but I understand that this might be particularly difficult.

kyliau:
I revisited the [end-to-end test](https://code.visualstudio.com/api/working-with-extensions/testing-extension) for vscode, and realized that we are relying on a timer to wait for the extension to be ready. 

https://github.com/angular/vscode-ng-language-service/blob/56176b210db79b90e0e71ce5e8c610d75bb6bbec/integration/e2e/helper.ts#L8-L18

If we use the same mechanism for ngcc, the test will be flaky since ngcc usually takes up a bit of time.
Our CircleCI setup also does not currently support such test (due to lack of windowing system).

I've created a [card](https://github.com/orgs/angular/projects/1#card-49562829) for myself to look into this more, but in the meantime I am going to proceed with this PR to get the ngcc work going.

ayazhafiz:
Ouch, yeah ngcc seems to take 15s+ whenever I run it. Maybe we can mock it, but I agree this is low priority

PR #980, opened 2020-11-18
build: update @angular/language-service to v11.1.0-next.0

kyliau:
Use `master` for the `next` version development, stable is now released
from `0.1100.x` branch

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -12,7 +12,7 @@
     "node": ">=10.9.0 <13.0.0"
   },
   "dependencies": {
-    "@angular/language-service": "11.0.1",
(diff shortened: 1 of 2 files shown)


ayazhafiz:
So is the release strategy now to have a separate branch for every major version?

kyliau:
We'll have a separate branch for every _minor_ version (or major if there's no next minor), so that we could use master for `next` and the stable branch for `latest`.

ayazhafiz:
Oh I see we're using semver. Okay sgtm

ayazhafiz (approved):

PR #981, opened 2020-11-19
fix: mark project as dirty after enabling language service

kyliau:
This commit fixes a bug in which the program becomes unavailable after
the language service is re-enabled, resulting in a crash.

When the language service is disabled, the program gets discarded.
However, when the language service is re-enabled, the program is not
recreated.

This bug was difficult to reproduce because it was masked by the fact that
during normal startup, ngcc gets run, and in the process, ngcc generates a
lockfile named `__ngcc_lock_file__` in the `node_modules` directory.
This triggers the directory watcher, and through a long chain of events,
reloads the Configured project, therefore re-creating the program.

Without ngcc run (make it a no-op), the extension will crash because the
language service is unable to return the program.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -6,6 +6,7 @@ set -ex -o pipefail
 (
   cd integration/project
   yarn
+  yarn ngcc
 )
 
 # Server unit tests
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -125,6 +125,11 @@ export class Session {
       return;
     }
     project.enableLanguageService();
+    // When the language service got disabled, the program was discarded via
+    // languageService.cleanupSemanticCache(). However, the program is not
+    // recreated when the language service is re-enabled. We manually mark the
+    // project as dirty to force update the graph.
+    project.markAsDirty();
     this.info(`Enabling Ivy language service for ${project.projectName}.`);
   }
 
--- a/integration/lsp/ivy_spec.ts
+++ b/integration/lsp/ivy_spec.ts
@@ -7,7 +7,7 @@
  */
 
 import {MessageConnection} from 'vscode-jsonrpc';
-
+import * as lsp from 'vscode-languageserver-protocol';
 import {APP_COMPONENT, createConnection, initializeServer, openTextDocument} from './test_utils';
 
 describe('Angular Ivy language server', () => {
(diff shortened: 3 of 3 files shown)


scripts/test.sh
```diff
@@ -6,6 +6,7 @@ set -ex -o pipefail
 (
   cd integration/project
   yarn
+  yarn ngcc
```
ayazhafiz, line 9: Hm.. I guess we need this, but this will slow down CI a bit.

server/src/session.ts
```diff
@@ -125,6 +125,11 @@ export class Session {
       return;
     }
     project.enableLanguageService();
+    // When the language service got disabled, the program was discarded via
+    // languageService.cleanupSemanticCache(). However, the program is not
+    // recreated when the language service is re-enabled. We manually mark the
+    // project as dirty to force update the graph.
+    project.markAsDirty();
```
ayazhafiz, line 132: I don't like all these "gotcha"s we keep running into with internal TS apis that are not well documented. Maybe another reason to migrate away from relying on the plugin model.

ayazhafiz (approved):

scripts/test.sh
kyliau: yeah, it's gonna take more time for CI to run, but as we add more tests for Ivy, this will become necessary.

server/src/session.ts
kyliau: Do you mean we rewrite the server ourselves? We need some kind of integration with TypeScript, but so far it is not clear to me where the boundary should be. Even if we don't use the tsserver plugin model, we still rely on tsserver for project management and document tracking. Is there an ideal architecture you have in mind?

ayazhafiz: For example to avoid this problem, if we avoid the plugin model, we can have a project and keep the tsLs for the project always enabled. Then we can enable/disable the angular language service as needed without having to go through the project, and the project will still update the tsLs language service (and hence the program) as needed. A potential concern is that the project may then be updating the tsLs language service more often than necessary, but I think in practice this will not be an issue.

kyliau: Ah I see what you mean - you're suggesting we maintain two distinct language service instances - tsserver maintains its default TypeScript language service, and we keep our own Angular language service. As it is right now, there is really not a way for us to distinguish one from another. From the project point of view, there is only one combined LS, and therefore one combined state. Yeah, I can see the two-fold benefits of separating them (distinct state + our own LS interface). Let's discuss this more when we meet in person 😄

PR #988, opened 2020-11-20
fix: match external templates to project on startup

kyliau:
Currently, Ivy LS is not able to match external projects to their external
templates if no TS files are open. This is because Ivy LS does not implement
`getExternalFiles()`.

To fix this, we take a different route:
After ngcc has run, we send diagnostics for all the open files. But in the
case where all open files are external templates, we first get diagnostics
for a root TS file, then process the rest. Processing a root TS file triggers
a global analysis, during which we match the external templates to their
project.

For a more detailed explanation, see
https://github.com/angular/vscode-ng-language-service/wiki/Project-Matching-for-External-Templates

### DIFF
--- a/integration/lsp/ivy_spec.ts
+++ b/integration/lsp/ivy_spec.ts
@@ -11,18 +11,19 @@ import * as lsp from 'vscode-languageserver-protocol';
 
 import {NgccComplete, ProjectLanguageService, ProjectLanguageServiceParams, RunNgcc, RunNgccParams} from '../../common/notifications';
 
-import {APP_COMPONENT, createConnection, initializeServer, openTextDocument} from './test_utils';
+import {APP_COMPONENT, createConnection, FOO_TEMPLATE, initializeServer, openTextDocument} from './test_utils';
 
 describe('Angular Ivy language server', () => {
   jasmine.DEFAULT_TIMEOUT_INTERVAL = 10000; /* 10 seconds */
 
   let client: MessageConnection;
 
-  beforeEach(() => {
+  beforeEach(async () => {
     client = createConnection({
       ivy: true,
     });
     client.listen();
+    await initializeServer(client);
   });
 
   afterEach(() => {
@@ -30,28 +31,24 @@ describe('Angular Ivy language server', () => {
   });
 
   it('should send ngcc notification after a project has finished loading', async () => {
-    await initializeServer(client);
     openTextDocument(client, APP_COMPONENT);
     const configFilePath = await onRunNgccNotification(client);
     expect(configFilePath.endsWith('integration/project/tsconfig.json')).toBeTrue();
   });
 
   it('should disable language service until ngcc has completed', async () => {
-    await initializeServer(client);
     openTextDocument(client, APP_COMPONENT);
     const languageServiceEnabled = await onLanguageServiceStateNotification(client);
     expect(languageServiceEnabled).toBeFalse();
   });
 
   it('should re-enable language service once ngcc has completed', async () => {
-    await initializeServer(client);
     openTextDocument(client, APP_COMPONENT);
     const languageServiceEnabled = await waitForNgcc(client);
     expect(languageServiceEnabled).toBeTrue();
   });
 
   it('should handle hover on inline template', async () => {
-    await initializeServer(client);
     openTextDocument(client, APP_COMPONENT);
     const languageServiceEnabled = await waitForNgcc(client);
     expect(languageServiceEnabled).toBeTrue();
@@ -65,6 +62,23 @@ describe('Angular Ivy language server', () => {
       language: 'typescript',
       value: '(property) AppComponent.name: string',
     });
+  });
+
+  it('should show existing diagnostics on external template', async () => {
+    client.sendNotification(lsp.DidOpenTextDocumentNotification.type, {
+      textDocument: {
+        uri: `file://${FOO_TEMPLATE}`,
+        languageId: 'typescript',
+        version: 1,
+        text: `{{ doesnotexist }}`,
+      },
+    });
+    const languageServiceEnabled = await waitForNgcc(client);
+    expect(languageServiceEnabled).toBeTrue();
+    const diagnostics = await getDiagnosticsForFile(client, FOO_TEMPLATE);
+    expect(diagnostics.length).toBe(1);
+    expect(diagnostics[0].message)
+        .toBe(`Property 'doesnotexist' does not exist on type 'FooComponent'.`);
   });
 });
 
@@ -81,6 +95,18 @@ function onLanguageServiceStateNotification(client: MessageConnection): Promise<
     client.onNotification(ProjectLanguageService, (params: ProjectLanguageServiceParams) => {
       resolve(params.languageServiceEnabled);
     });
+  });
+}
+
+function getDiagnosticsForFile(
+    client: MessageConnection, fileName: string): Promise<lsp.Diagnostic[]> {
+  return new Promise(resolve => {
+    client.onNotification(
+        lsp.PublishDiagnosticsNotification.type, (params: lsp.PublishDiagnosticsParams) => {
+          if (params.uri === `file://${fileName}`) {
+            resolve(params.diagnostics);
+          }
+        });
   });
 }
 
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -15,7 +15,7 @@ import * as notification from '../common/notifications';
 import {tsCompletionEntryToLspCompletionItem} from './completion';
 import {tsDiagnosticToLspDiagnostic} from './diagnostic';
 import {ServerHost} from './server_host';
-import {filePathToUri, lspPositionToTsPosition, lspRangeToTsPositions, tsTextSpanToLspRange, uriToFilePath} from './utils';
+import {filePathToUri, isConfiguredProject, lspPositionToTsPosition, lspRangeToTsPositions, tsTextSpanToLspRange, uriToFilePath} from './utils';
 
 export interface SessionOptions {
   host: ServerHost;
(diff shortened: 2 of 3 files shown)


server/src/session.ts
```diff
@@ -131,6 +131,51 @@ export class Session {
     // project as dirty to force update the graph.
     project.markAsDirty();
     this.info(`Enabling Ivy language service for ${project.projectName}.`);
+
+    // Send diagnostics since we skipped this step when opening the file
+    // (because language service was disabled while waiting for ngcc).
+    // Technically we could just call project.refreshDiagnostics(), but we need
+    // to handle the special case where all open files are external templates.
+    const openFiles = this.getOpenFilesForNewlyLoadedProject(project);
```
ayazhafiz, line 139: I think it would be useful to explain the entire strategy somewhere, maybe in the doc comment for this method. If I recall correctly: when someone opens an external template, the project service walks directories and finds all projects, for all projects we check if it is an Ng project and run ngcc, then we get back the ngcc complete response, and at this point we get trigger diagnostics for a typescript file in the project, which would cause an incremental build of the compiler, in the process discovering all external templates and attaching them to the project. (More to make sure I remember this :) )

server/src/session.ts
```diff
@@ -131,6 +131,51 @@ export class Session {
...
+   * Get the open files for a newly loaded project. If all the open files are
+   * external templates, then try to match them to the specified `project` by
+   * including one of the root files. The inclusion of a root file triggers
+   * global analysis, during which we fix up the external templates.
+   */
+  private getOpenFilesForNewlyLoadedProject(project: ts.server.Project): string[] {
+    const openFiles: string[] = [];
+
+    const iterable = this.projectService.openFiles.keys();
+    for (let it = iterable.next(); !it.done; it = iterable.next()) {
```
ayazhafiz, line 153: Does `for..of` not work here?

server/src/session.ts
```diff
@@ -131,6 +131,51 @@ export class Session {
...
+      const fileName = it.value;
+      const scriptInfo = this.projectService.getScriptInfo(fileName);
+      // External templates without a project are orphans.
+      if (scriptInfo?.isOrphan() || scriptInfo?.attachToProject(project)) {
+        openFiles.push(scriptInfo.fileName);
+      }
```
ayazhafiz, line 159: At this point, shouldn't all external templates be orphans? What is the significance of `scriptInfo?.attachToProject(project)`?

server/src/session.ts
```diff
@@ -131,6 +131,51 @@ export class Session {
...
+    }
+
+    if (!openFiles.some(isTypeScriptFile)) {
+      if (!project.hasRoots()) {
+        // If all open files are external templates and there is no root file,
+        // no point getting diagnostics since we won't be able to return any
+        // information. This is extremely unlikely.
```
ayazhafiz, line 166: Can you add a comment explaining why this is the case?

server/src/session.ts
```diff
@@ -131,6 +131,51 @@ export class Session {
...
+    this.sendPendingDiagnostics(openFiles);
+  }
+
+  /**
+   * Get the open files for a newly loaded project. If all the open files are
+   * external templates, then try to match them to the specified `project` by
+   * including one of the root files. The inclusion of a root file triggers
+   * global analysis, during which we fix up the external templates.
+   */
+  private getOpenFilesForNewlyLoadedProject(project: ts.server.Project): string[] {
```
ayazhafiz, line 149: What do you think of splitting this into two steps/functions:

1) Get a single TS file to perform global analysis, and run `getDiagnostics` for it to do the global analysis
2) Return diagnostics for all open files

I think this would be a little easier to reason about and should be the same performance-wise.

server/src/session.ts
kyliau: This is [TS-internal iterator](https://github.com/microsoft/TypeScript/blob/eac0f6dc3085647ce301bb79516cd925e528911c/src/compiler/corePublic.ts#L90), which is different from the ES 2015 iterator protocol. It does not have the `iterator` symbol, so `for..of` doesn't work. I'm not quite sure why TS needs its own iterator interface .. perhaps for compatibility with older runtimes that do not support ES 2015? 🤷‍♂️ 
They even have their own `arrayFrom`, but it's an internal method:
https://github.com/microsoft/TypeScript/blob/eac0f6dc3085647ce301bb79516cd925e528911c/src/compiler/core.ts#L1329-L1337

server/src/session.ts
kyliau: Line 157 here is wrong for two reasons (shouldn't have written code when I'm sleepy):
1. all the script infos are not actually orphans, because by default they get attached to an `inferred` project. 
    What we want to check here is really "if the script info is not attached to any configured project"
2. I meant to call `isAttachedTo(project)`, not `attachToProject()` 😅

server/src/session.ts
kyliau: I made some minor revisions to the same doc I wrote and posted it to https://github.com/angular/vscode-ng-language-service/wiki/Project-Matching-for-External-Templates
I've also added the link to the commit and PR description.
Your description of the entire workflow is spot-on!

server/src/session.ts
kyliau: This is excellent suggestion! I've simplified the whole logic, albeit at a slight inefficiency -
Previously, we look at all open files, check if there is at least one TS file, and if not inject one of the root files.
With the latest change, we do (1) unconditionally, then just refresh diagnostics for the whole project.

server/src/session.ts
ayazhafiz: Odd, thanks for the details!

integration/lsp/ivy_spec.ts
```diff
@@ -63,6 +61,30 @@ describe('Angular Ivy language server', () => {
...
+    const languageServiceEnabled = await waitForNgcc(client);
+    expect(languageServiceEnabled).toBeTrue();
+    const diagnostics: lsp.Diagnostic[] = await new Promise(resolve => {
+      client.onNotification(
+          lsp.PublishDiagnosticsNotification.type, (params: lsp.PublishDiagnosticsParams) => {
+            if (params.uri === `file://${FOO_TEMPLATE}`) {
+              resolve(params.diagnostics);
+            }
+          });
+    });
```
ayazhafiz, line 83: nit: consider making a helper function

server/src/session.ts
```diff
@@ -131,6 +131,24 @@ export class Session {
     // project as dirty to force update the graph.
     project.markAsDirty();
     this.info(`Enabling Ivy language service for ${project.projectName}.`);
+
+    // Send diagnostics since we skipped this step when opening the file
+    // (because language service was disabled while waiting for ngcc).
+    this.runGlobalAnalysisForNewlyLoadedProject(project);
+    project.refreshDiagnostics();  // Show initial diagnostics
```
ayazhafiz, line 138:
```suggestion
    // Send diagnostics since we skipped this step when opening the file
    // (because language service was disabled while waiting for ngcc).
    // First, make sure the Angular project is complete.
    this.runGlobalAnalysisForNewlyLoadedProject(project);
    project.refreshDiagnostics();  // Show initial diagnostics
```

server/src/session.ts
```diff
@@ -131,6 +131,24 @@ export class Session {
...
+  }
+
+  /**
+   * Invoke the compiler for the first time so that external templates get
+   * matched to the project they belong to.
+   */
+  private runGlobalAnalysisForNewlyLoadedProject(project: ts.server.Project) {
+    if (!project.hasRoots()) {
+      return;
+    }
```
ayazhafiz, line 148: Can you add a comment that this is case means we have to wait until a TS file manually opened, but is very rare, as you did before?

server/src/session.ts
```diff
@@ -15,7 +15,7 @@ import {tsCompletionEntryToLspCompletionItem} from './completion';
 import {tsDiagnosticToLspDiagnostic} from './diagnostic';
 import {Logger} from './logger';
 import {ServerHost} from './server_host';
-import {filePathToUri, lspPositionToTsPosition, lspRangeToTsPositions, tsTextSpanToLspRange, uriToFilePath} from './utils';
+import {filePathToUri, isConfiguredProject, isTypeScriptFile, lspPositionToTsPosition, lspRangeToTsPositions, tsTextSpanToLspRange, uriToFilePath} from './utils';
```
ayazhafiz, line 18: Looks like `isTypeScriptFile` is never used in this file, can we remove it entirely?

integration/lsp/ivy_spec.ts
kyliau: done

server/src/session.ts
kyliau: removed

server/src/session.ts
kyliau: Could you please elaborate more?
In the previous comment, I meant "a project without any root files" is extremely rare. It could be that either the project indeed does not have any files, or something went wrong in the loading phase such that the project is still empty. The second case should not happen at all, and in the first case, there isn't any file that the user could open that'd affect this project.

server/src/session.ts
```diff
@@ -131,6 +131,25 @@ export class Session {
...
+  /**
+   * Invoke the compiler for the first time so that external templates get
+   * matched to the project they belong to.
+   */
+  private runGlobalAnalysisForNewlyLoadedProject(project: ts.server.Project) {
+    if (!project.hasRoots()) {
+      return;
+    }
+    const fileName = project.getRootScriptInfos()[0].fileName;
+    // Getting semantic diagnostics will trigger a global analysis.
```
atscott, line 151: Is it that getting semantic diagnostics always triggers global analysis or is this just because we're triggering diagnostics on a root file?

atscott (approved):

server/src/session.ts
kyliau: Getting semantic diagnostics always triggers global analysis, because a new compiler is created every time. Technically any file will do, not just root files, but the file must be part of the project, so root file is a convenient candidate.

server/src/session.ts
kyliau: @ayazhafiz could you please take another look and let me know if there's further clarification needed?

ayazhafiz: No more clarification needed, LGTM :)

ayazhafiz (approved):

PR #992, opened 2020-11-20
refactor: remove `traceToConsole` feature in Logger

kyliau:
Currently if users ask for logging to disabled (by not providing a filename
for the log file), the Logger defaults to writing to console.

In practice, this feature is not used at all. If users do not want logging,
then we should just disable logging entirely.

### DIFF
--- a/server/src/logger.ts
+++ b/server/src/logger.ts
@@ -38,7 +38,7 @@ export interface LoggerOptions {
  * Create a logger instance to write to file.
  * @param options Logging options.
  */
-export function createLogger(options: LoggerOptions): Logger {
+export function createLogger(options: LoggerOptions): ts.server.Logger {
   let logLevel: ts.server.LogLevel;
   switch (options.logVerbosity) {
     case 'requestTime':
@@ -55,9 +55,7 @@ export function createLogger(options: LoggerOptions): Logger {
       logLevel = ts.server.LogLevel.terse;
       break;
   }
-  // If logFile is not provided then just trace to console.
-  const traceToConsole = !options.logFile;
-  return new Logger(traceToConsole, logLevel, options.logFile);
+  return new Logger(logLevel, options.logFile);
 }
 
 // TODO: Code below is from TypeScript's repository. Maybe create our own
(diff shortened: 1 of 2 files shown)


atscott (approved):

server/src/logger.ts
```diff
@@ -132,15 +125,15 @@ export class Logger implements ts.server.Logger {
...
+    return this.fd >= 0;
   }
 
   hasLevel(level: ts.server.LogLevel) {
     return this.loggingEnabled() && this.level >= level;
   }
 
   msg(s: string, type: ts.server.Msg = ts.server.Msg.Err) {
-    if (!this.canWrite) return;
+    if (!this.loggingEnabled()) return;
```
ayazhafiz, line 136: Check is unnecessary since `write` checks for this already

server/src/logger.ts
```diff
@@ -72,14 +70,13 @@ function nowString() {
   return `${d.getHours()}:${d.getMinutes()}:${d.getSeconds()}.${d.getMilliseconds()}`;
 }
 
-export class Logger implements ts.server.Logger {
-  private fd = -1;
+class Logger implements ts.server.Logger {
+  private readonly fd: number = -1;
```
ayazhafiz, line 74: Should this always be `-1`? Then `loggingEnabled` will never be true.

kyliau: `this.fd` is updated on line 89 if a filename is provided

server/src/logger.ts
kyliau: yeah, but with this check we can be slightly more efficient by not processing the string unnecessarily

server/src/logger.ts
ayazhafiz: Can we set -1 there instead?

server/src/logger.ts
ayazhafiz: Can we just move the contents of `write` into here then, as it's the only caller

ayazhafiz (approved):

server/src/logger.ts
kyliau: done

server/src/logger.ts
kyliau: done

PR #994, opened 2020-11-23
build: output all build artifacts to dist directory

kyliau:
This commit refactors the build architecture to output all artifacts to
the `dist` directory. With this change, we make the repository a monorepo
so that code in `common` directory can be used seamlessly in `client` and
`server`, both in production source code and test (not possible before).

As a side effect of this, due to the deduplication of dependencies between `client`
and `server`, we are able to reduce the number of files in the final `.vsix` bundle
from 478 files (10.99mb) down to 393 files (10.82mb) and get rid of the following warning:

> This extension consists of 478 files, out of which 127 are JavaScript files. For performance reasons, you should bundle your extension: https://aka.ms/vscode-bundle-extension . You should also exclude unnecessary files by adding them to your .vscodeignore: https://aka.ms/vscode-vscodeignore

The downside of a monorepo is that the dependencies of the `server` need to be duplicated in the
root `package.json`. Therefore, if a new prod dependency is added to the server, it needs to be
added to both `package.json` and `server/package.json`.

View of the final `npm` directory (what gets published to npm and the vscode marketplace):
```shell
$ tree dist/npm -I node_modules
dist/npm
├── CHANGELOG.md
├── README.md
├── angular.png
├── index.js
├── package.json
├── server
│   ├── README.md
│   ├── index.js
│   └── package.json
├── syntaxes
│   ├── expression.json
│   ├── inline-styles.json
│   ├── inline-template.json
│   └── template.json
└── yarn.lock
```


ayazhafiz (commented):
Looks like we could reduce the size of the published extension package by taking advantage of the `.vscodeignore` mentioned in your quoted message.

ayazhafiz (approved):

kyliau:
> Looks like we could reduce the size of the published extension package by taking advantage of the `.vscodeignore` mentioned in your quoted message.

If you look at the final structure in the PR description, our bundle is pretty optimized actually. There are only 13 files. The rest of the files come from `node_modules`. In order to further minimize the bundle size and number of files, we'll have to run rollup over external dependencies, like `vscode-languageclient`, `vscode-languageserver`, etc.

ayazhafiz:
Oh I was thinking of dropping the markdown files, image, etc. But I just realized you need that for the extension to render nicely in the marketplace.

DEVELOPER.md
```diff
@@ -1,3 +1,17 @@
...
+The `dependencies` in root [`package.json`](package.json) must include all production dependencies
+of the client **and** and the server.
+
+The dependencies of the `server` need to be duplicated in the root [`package.json`](package.json).
+Therefore, if a new prod dependency is added to the server, it needs to be
+added to both [`package.json`](package.json) and [`server/package.json`](server/package.json).
+
+Note that we do not install the dependencies in [`server`](server),
+hence there should not be a `yarn.lock` file.
+
```
atscott, line 14: Does all of the content in the section below still make sense? It was added in the initial commit for the common code: https://github.com/angular/vscode-ng-language-service/commit/fee9166ebeed9291220471983464a935b222d4ae
I think this might change that a little bit? Or do you still need to run `yarn compile`?

atscott (approved):

DEVELOPER.md
kyliau: The content in the section below still stands, because we are still using TypeScript's project references for the build.
Moreover, if you look at `client/tsconfig.json` and `server/tsconfig.json`, the `dist` directory is one of the `rootDirs`, so code in this repo must be compiled to enable project navigation in IDE.

PR #995, opened 2020-11-24
feat: add getReferencesAtPosition to capabilities

atscott:
The Angular Language Service for Ivy will have a be able to provide
references for a template location so we need to add `referencesProvider` to the capabilities.

### DIFF
--- a/integration/lsp/viewengine_spec.ts
+++ b/integration/lsp/viewengine_spec.ts
@@ -124,6 +124,7 @@ describe('initialization', () => {
         definitionProvider: true,
         typeDefinitionProvider: false,
         hoverProvider: true,
+        referencesProvider: false,
         workspace: {
           workspaceFolders: {
             supported: true,
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -103,6 +103,7 @@ export class Session {
     conn.onDidSaveTextDocument(p => this.onDidSaveTextDocument(p));
     conn.onDefinition(p => this.onDefinition(p));
     conn.onTypeDefinition(p => this.onTypeDefinition(p));
+    conn.onReferences(p => this.onReferences(p));
     conn.onHover(p => this.onHover(p));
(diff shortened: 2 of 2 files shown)


server/src/session.ts
```diff
@@ -386,6 +388,25 @@ export class Session {
...
 
+  private onReferences(params: lsp.TextDocumentPositionParams): lsp.Location[]|undefined {
+    const lsInfo = this.getLSAndScriptInfo(params.textDocument);
+    if (lsInfo === undefined) {
+      return;
+    }
+    const {languageService, scriptInfo} = lsInfo;
+    const offset = lspPositionToTsPosition(scriptInfo, params.position);
+    const references = languageService.getReferencesAtPosition(scriptInfo.fileName, offset);
+    if (!references) {
```
kyliau, line 399: prefer explicit check? I don't think we have an official style guide yet, do we? 
In the meantime we should at least be consistent (compare line 393).

kyliau (approved):

integration/lsp/viewengine_spec.ts
```diff
@@ -124,6 +124,7 @@ describe('initialization', () => {
         definitionProvider: true,
         typeDefinitionProvider: true,
         hoverProvider: true,
+        referencesProvider: true,
```
kyliau, line 127: Actually, now that we have the `ivy` flag, we should enable this only for Ivy

atscott: Good call. Done

server/src/session.ts
atscott: Oops, yes. I often write these like this as an initial pass because it's less typing >.>

kyliau (approved):

kyliau:
@atscott could you please rebase and merge?

atscott:
> @atscott could you please rebase and merge?

@kyliau Done. Thanks for the ping

PR #1003, opened 2020-12-01
build: update rollup plugin and fix build warnings

kyliau:
Two commits:

1. Use @rollup/commonjs-plugin
2. Fix all build warnings about external deps when bundling extension and server

### DIFF
--- a/banner.rollup.config.js
+++ b/banner.rollup.config.js
@@ -6,6 +6,7 @@ module.exports = [
       format: 'cjs',
     },
     external: [
+      'fs',
       'path',
     ],
   },
--- a/package.json
+++ b/package.json
@@ -136,7 +136,7 @@
     "jasmine": "3.6.3",
     "prettier": "2.2.1",
     "rollup": "2.34.0",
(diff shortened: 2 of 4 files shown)


ayazhafiz (approved):

ayazhafiz:
Test failure seems legitimate, can you re-request review when fixed? Maybe removal of `'conditional-runtime-dependency'` from the ignore broke it?

kyliau:
@ayazhafiz yeah you're right, something is off about dynamic require. I removed `conditional-runtime-dependency` because that option is removed from the new plugin. It is replaced by [`dynamicRequireTargets`](https://github.com/rollup/plugins/tree/master/packages/commonjs#dynamicrequiretargets). Adding the new option fixed CI, but increased bundle size from 10.82MB to 11.53MB.
I'll investigate more.

kyliau:
Look like `rollup/commonjs-plugin` is not very stable yet, I'm running into this issue: https://github.com/rollup/plugins/issues/609
Will park this for now and revisit after Ivy LS.

PR #1007, opened 2020-12-04
build: use NPM to install node_modules

kyliau:
`semver` is a prod dependency of `vscode-languageclient`.
In the old build pipeline, `vscode-languageclient` is declared in
`client/package.json`.
When yarn installs dependencies in the `client` directory, it will put `semver`
next to `vscode-languageclient` in `client/node_modules`.

In the new build pipeline, `vscode-languageclient` is declared in root
`package.json`.
When yarn installs dependencies in the root directory, it chooses not to hoist
`semver`, instead puts it in a nested node_modules under
`node_modules/vscode-languageclient/node_modules`.

There is a bug in vsce that causes it to ignore nested `node_modules`,
resulting in a `vsix` build that is broken. The issue is tracked
[here](https://github.com/microsoft/vscode-vsce/issues/432), but it does not look like a fix will come any time soon.

For now we should install our dependencies using NPM.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -13,7 +13,7 @@ rm -rf **/*.tsbuildinfo
 yarn run compile
 
 # Copy files to package root
-cp package.json yarn.lock angular.png CHANGELOG.md README.md dist/npm
+cp package.json angular.png CHANGELOG.md README.md dist/npm
 # Copy files to server directory
 cp server/package.json server/README.md dist/npm/server
 # Build and copy files to syntaxes directory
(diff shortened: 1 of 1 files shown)


ayazhafiz (approved):
Thanks for the great explanation.

scripts/build.sh
```diff
@@ -23,9 +23,9 @@ mkdir dist/npm/syntaxes
 cp syntaxes/!(tsconfig).json dist/npm/syntaxes
 
 pushd dist/npm
-yarn install --production --ignore-scripts
```
atscott, line 26: Would it be useful to add a TODO comment here that either links to the `vsce` issue or this PR so we know to revert this when the issue is fixed?

atscott (approved):

scripts/build.sh
kyliau: Good call! Added TODO

PR #1008, opened 2020-12-04
Command to view template typecheck block

ayazhafiz:
This patch adds a command to retrieve and display the typecheck block
for a template under the user's active selection (if any), and
highlights the span of the node(s) in the typecheck block that
correspond to the template node under the user's active selection (if
any). The typecheck block is made available via a dedicated text
document provider that queries fresh typecheck block content whenever
the `getTemplateTcb` command is invoked.

Probably not something we want to land soon, but a useful debugging tool
for folks working with TCBs.

See also https://github.com/angular/angular/pull/39974, which provides
the language service implementations needed for this feature.

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -9,14 +9,20 @@
 import * as vscode from 'vscode';
 import {ServerOptions} from '../common/initialize';
 import {AngularLanguageClient} from './client';
+import {ANGULAR_SCHEME, TcbContentProvider} from './providers';
 
 /**
  * Represent a vscode command with an ID and an impl function `execute`.
  */
-interface Command {
-  id: string;
-  execute(): Promise<unknown>;
-}
+type Command = {
+  id: string,
+  isTextEditorCommand: false,
+  execute(): Promise<unknown>,
+}|{
+  id: string,
+  isTextEditorCommand: true,
+  execute(textEditor: vscode.TextEditor): Promise<unknown>,
+};
 
 /**
  * Restart the language server by killing the process then spanwing a new one.
@@ -26,6 +32,7 @@ interface Command {
 function restartNgServer(client: AngularLanguageClient): Command {
   return {
     id: 'angular.restartNgServer',
+    isTextEditorCommand: false,
     async execute() {
       await client.stop();
       await client.start();
@@ -39,6 +46,7 @@ function restartNgServer(client: AngularLanguageClient): Command {
 function openLogFile(client: AngularLanguageClient): Command {
   return {
     id: 'angular.openLogFile',
+    isTextEditorCommand: false,
     async execute() {
       const serverOptions: ServerOptions|undefined = client.initializeResult?.serverOptions;
       if (!serverOptions?.logFile) {
@@ -65,6 +73,50 @@ function openLogFile(client: AngularLanguageClient): Command {
 }
 
 /**
+ * Command getTemplateTcb displays a typecheck block for the template a user has
+ * an active selection over, if any.
+ * @param ngClient LSP client for the active session
+ * @param context extension context to which disposables are pushed
+ */
+function getTemplateTcb(
+    ngClient: AngularLanguageClient, context: vscode.ExtensionContext): Command {
+  const TCB_HIGHLIGHT_DECORATION = vscode.window.createTextEditorDecorationType({
+    // See https://code.visualstudio.com/api/references/theme-color#editor-colors
+    backgroundColor: new vscode.ThemeColor('editor.selectionHighlightBackground'),
+  });
+
+  const tcbProvider = new TcbContentProvider();
+  const disposable = vscode.workspace.registerTextDocumentContentProvider(
+      ANGULAR_SCHEME,
+      tcbProvider,
+  );
+  context.subscriptions.push(disposable);
+
+  return {
+    id: 'angular.getTemplateTcb',
+    isTextEditorCommand: true,
+    async execute(textEditor: vscode.TextEditor) {
+      tcbProvider.clear();
+      const response = await ngClient.getTcbUnderCursor(textEditor);
+      if (response === undefined) {
+        return undefined;
+      }
+      // Change the scheme of the URI from `file` to `ng` so that the document
+      // content is requested from our own `TcbContentProvider`.
+      const tcbUri = response.uri.with({
+        scheme: ANGULAR_SCHEME,
+      });
+      tcbProvider.update(tcbUri, response.content);
+      const editor = await vscode.window.showTextDocument(tcbUri, {
+        viewColumn: vscode.ViewColumn.Beside,
+        preserveFocus: true,  // cursor remains in the active editor
+      });
+      editor.setDecorations(TCB_HIGHLIGHT_DECORATION, response.selections);
+    }
+  };
+}
+
+/**
  * Register all supported vscode commands for the Angular extension.
  * @param client language client
  * @param context extension context for adding disposables
@@ -74,10 +126,13 @@ export function registerCommands(
   const commands: Command[] = [
     restartNgServer(client),
     openLogFile(client),
+    getTemplateTcb(client, context),
   ];
 
   for (const command of commands) {
-    const disposable = vscode.commands.registerCommand(command.id, command.execute);
+    const disposable = command.isTextEditorCommand ?
+        vscode.commands.registerTextEditorCommand(command.id, command.execute) :
+        vscode.commands.registerCommand(command.id, command.execute);
     context.subscriptions.push(disposable);
   }
 }
--- a/client/src/providers.ts
+++ b/client/src/providers.ts
@@ -1,0 +1,62 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as vscode from 'vscode';
+
+export const ANGULAR_SCHEME = 'ng';
+
+/**
+ * Allocate a provider of documents corresponding to the `ng` URI scheme,
+ * which we will use to provide a virtual document with the TCB contents.
+ *
+ * We use a virtual document provider rather than opening an untitled file to
+ * ensure the buffer remains readonly (https://github.com/microsoft/vscode/issues/4873).
+ */
+export class TcbContentProvider implements vscode.TextDocumentContentProvider {
+  /**
+   * Event emitter used to notify VSCode of a change to the TCB virtual document,
+   * prompting it to re-evaluate the document content. This is needed to bust
+   * VSCode's document cache if someone requests a TCB that was previously opened.
+   * https://code.visualstudio.com/api/extension-guides/virtual-documents#update-virtual-documents
+   */
+  private readonly onDidChangeEmitter = new vscode.EventEmitter<vscode.Uri>();
+  /**
+   * Name of the typecheck file.
+   */
+  private tcbFile: vscode.Uri|null = null;
+  /**
+   * Content of the entire typecheck file.
+   */
+  private tcbContent: string|null = null;
+
+  /**
+   * This callback is invoked only when user explicitly requests to view or
+   * update typecheck file. We do not automatically update the typecheck document
+   * when the source file changes.
+   */
+  readonly onDidChange = this.onDidChangeEmitter.event;
+
+  provideTextDocumentContent(uri: vscode.Uri, token: vscode.CancellationToken):
+      vscode.ProviderResult<string> {
+    if (uri.toString() !== this.tcbFile?.toString()) {
+      return null;
+    }
+    return this.tcbContent;
+  }
+
+  update(uri: vscode.Uri, content: string) {
+    this.tcbFile = uri;
+    this.tcbContent = content;
+    this.onDidChangeEmitter.fire(uri);
+  }
+
+  clear() {
+    this.tcbFile = null;
+    this.tcbContent = null;
+  }
+}
--- a/package.json
+++ b/package.json
@@ -27,8 +27,22 @@
         "command": "angular.openLogFile",
         "title": "Open Angular Server log",
         "category": "Angular"
+      },
+      {
+        "command": "angular.getTemplateTcb",
+        "title": "View Template Typecheck Block",
+        "category": "Angular"
       }
     ],
+    "menus": {
+      "editor/context": [
+        {
+          "when": "resourceLangId == html || resourceLangId == typescript",
+          "command": "angular.getTemplateTcb",
+          "group": "angular"
+        }
+      ]
+    },
     "configuration": {
       "title": "Angular Language Service",
       "properties": {
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -6,12 +6,14 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {NgLanguageService} from '@angular/language-service';
 import * as ts from 'typescript/lib/tsserverlibrary';
 import * as lsp from 'vscode-languageserver/node';
 
 import {ServerOptions} from '../common/initialize';
 import {ProjectLanguageService, ProjectLoadingFinish, ProjectLoadingStart, SuggestIvyLanguageService, SuggestStrictMode} from '../common/notifications';
 import {NgccProgressToken, NgccProgressType} from '../common/progress';
+import {GetTcbParams, GetTcbRequest, GetTcbResponse} from '../common/requests';
 
 import {readNgCompletionData, tsCompletionEntryToLspCompletionItem} from './completion';
 import {tsDiagnosticToLspDiagnostic} from './diagnostic';
@@ -132,6 +134,30 @@ export class Session {
     conn.onHover(p => this.onHover(p));
     conn.onCompletion(p => this.onCompletion(p));
     conn.onCompletionResolve(p => this.onCompletionResolve(p));
+    conn.onRequest(GetTcbRequest, p => this.onGetTcb(p));
+  }
+
+  private onGetTcb(params: GetTcbParams): GetTcbResponse|undefined {
+    const lsInfo = this.getLSAndScriptInfo(params.textDocument);
+    if (lsInfo === undefined) {
+      return undefined;
+    }
+    const {languageService, scriptInfo} = lsInfo;
+    const offset = lspPositionToTsPosition(scriptInfo, params.position);
+    const response = languageService.getTcb(scriptInfo.fileName, offset);
+    if (response === undefined) {
+      return undefined;
+    }
+    const {fileName: tcfName} = response;
+    const tcfScriptInfo = this.projectService.getScriptInfo(tcfName);
+    if (!tcfScriptInfo) {
+      return undefined;
+    }
+    return {
+      uri: filePathToUri(tcfName),
+      content: response.content,
+      selections: response.selections.map((span => tsTextSpanToLspRange(tcfScriptInfo, span))),
+    };
   }
 
   private async runNgcc(configFilePath: string) {
@@ -663,7 +689,7 @@ export class Session {
   }
 
   private getLSAndScriptInfo(textDocumentOrFileName: lsp.TextDocumentIdentifier|string):
-      {languageService: ts.LanguageService, scriptInfo: ts.server.ScriptInfo}|undefined {
+      {languageService: NgLanguageService, scriptInfo: ts.server.ScriptInfo}|undefined {
     const filePath = lsp.TextDocumentIdentifier.is(textDocumentOrFileName) ?
         uriToFilePath(textDocumentOrFileName.uri) :
         textDocumentOrFileName;
@@ -683,7 +709,7 @@ export class Session {
       return undefined;
     }
     return {
-      languageService: project.getLanguageService(),
+      languageService: project.getLanguageService() as NgLanguageService,
       scriptInfo,
     };
   }
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -13,8 +13,15 @@ import * as lsp from 'vscode-languageclient/node';
 
 import {ProjectLoadingFinish, ProjectLoadingStart, SuggestIvyLanguageService, SuggestIvyLanguageServiceParams, SuggestStrictMode, SuggestStrictModeParams} from '../common/notifications';
 import {NgccProgress, NgccProgressToken, NgccProgressType} from '../common/progress';
+import {GetTcbRequest} from '../common/requests';
 
 import {ProgressReporter} from './progress-reporter';
+
+interface GetTcbResponse {
+  uri: vscode.Uri;
+  content: string;
+  selections: vscode.Range[];
+}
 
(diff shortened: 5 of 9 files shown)


ayazhafiz:
cc @kyliau if you are interested in the `TextDocumentProvider` implementation

ayazhafiz:
Blocked on a release of angular containing https://github.com/angular/angular/commit/d482f5cdd3843b9b6782f17f3dbb7af660f400dd

kyliau:
@ayazhafiz `@angular/language-service` v11.2.0-next.0 is out, we can proceed with this PR now :)

ayazhafiz:
@kyliau sounds good, I will clean this up soon.

client/src/commands/getTcb.ts
```diff
@@ -0,0 +1,104 @@
...
+  // We immediately dispose of the provider so we can create a fresh provider
+  // for the `ng` scheme next time `getTcb` is requested.
+  disposeNgProvider.dispose();
+}
+
+/**
+ * Command getTemplateTcb displays a typecheck block for the template a user has
+ * an active selection over, if any.
+ */
+export function getTemplateTcb(ngClient: AngularLanguageClient) {
```
ayazhafiz, line 96: I could find no way to test the client-side behavior on the VSCode docs. Let me know if you have any ideas; for now I recommend just spinning up a session and see what happens. On my side it looked to be okay, I can add a demo if folks desire.

server/src/session.ts
```diff
@@ -121,6 +131,22 @@ export class Session {
...
+  private onGetTcb(params: lspext.GetTcbParams): ngls.GetTcbResponse {
+    const lsInfo = this.getLSAndScriptInfo(params.textDocument);
+    if (lsInfo === undefined) {
+      return undefined;
+    }
+    const {languageService, scriptInfo} = lsInfo;
+    const offset = lspPositionToTsPosition(scriptInfo, params.position);
+    const ls = languageService as ViewEngineOrIvyLanguageService;
+    if ('getTcb' in ls) {
+      return ls.getTcb(scriptInfo.fileName, offset);
```
atscott, line 146: Typically, the way this was done in the past would be add a stub implementation of `getTcb` in the VE LS that's just `return undefined`. I would prefer that a bit more since it requires less change when we eventually remove the VE version. 
With that, I think you can change the return type of `getLSAndScriptInfo` to be `NgLanguageService`

client/src/commands/getTcb.ts
```diff
@@ -0,0 +1,107 @@
...
+  backgroundColor: new vscode.ThemeColor('editor.selectionHighlightBackground'),
+});
+
+class SourceMap {
+  private readonly lines: Array<readonly[/* line offset */ number, /* line length */ number]>;
+  constructor(source: string) {
+    let offset = 0;
+    this.lines = source.split('\n').map(line => {
+      const lineLen = line.length + 1 /* ='\n' */;
+      const offsetAndLen = [offset, lineLen] as const;
```
atscott, line 24: Small nit on naming: my brain reads these as `lineLength` and `offsetLength` but it takes longer because it has to always do the conversion. I would personally prefer the words spelled out but I'll survive if you keep it as-is :)

package.json
```diff
@@ -27,6 +27,11 @@
         "command": "angular.openLogFile",
         "title": "Open Angular Server log",
         "category": "Angular"
+      },
+      {
+        "command": "angular.getTemplateTcb",
```
atscott, line 32: I wonder if it would be good to provide this command in the context menu of html files: https://stackoverflow.com/a/41400631

atscott (changes requested):

atscott:
This is _really_ cool by the way; thanks for taking this on!

client/src/commands/getTcb.ts
```diff
@@ -0,0 +1,107 @@
...
+      offset += lineLen;
+      return offsetAndLen;
+    });
+  }
+
+  toPosition(offset: number): vscode.Position {
```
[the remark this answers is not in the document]
ayazhafiz: it's the character offset. jsdoc added

client/src/commands/getTcb.ts
```diff
@@ -0,0 +1,107 @@
...
+    const lineno = this.lines.findIndex(
+        ([lineOffset, lineLen]) => offset >= lineOffset && offset < lineOffset + lineLen);
+    if (lineno === -1) throw new Error(`${offset} is out of bounds`);
+    const [line_offset, ] = this.lines[lineno];
+    return new vscode.Position(lineno, offset - line_offset);
+  }
+}
+
+async function executeGetTemplateTcb(
+    ngClient: AngularLanguageClient, didRequestTcb: vscode.EventEmitter<vscode.Uri>) {
```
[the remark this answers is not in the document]
ayazhafiz: It's a void function, annotated as such.

server/src/session.ts
ayazhafiz: Yeah you're right. We'll update on the angular side.

client/src/commands/getTcb.ts
ayazhafiz: I am not sure I follow, but I tried to rename the variables to spell everything out. Can you let me know if it is better?

package.json
ayazhafiz: neat idea! done.

package.json
```diff
@@ -109,7 +114,16 @@
         "path": "./syntaxes/expression.json",
         "scopeName": "expression.ng"
       }
-    ]
+    ],
+    "menus": {
```
atscott, line 118: Because these items map directly to commands that are defined above, does it make sense to keep the "commands" item and the "menus" item next to each other?

atscott (approved):

package.json
ayazhafiz: sure!

client/src/commands.ts
```diff
@@ -74,6 +75,7 @@ export function registerCommands(
   const commands: Command[] = [
     restartNgServer(client),
     openLogFile(client),
+    getTemplateTcb(client),
```
atscott, line 78: One thing I noticed here is that the other commands in this list don't depend on the Angular LS being initialized yet, but this one does. We might want to consider _not_ registering this command right away, but instead keep a separate list of commands that we register once the initialization is complete.

package.json
```diff
@@ -27,8 +27,22 @@
...
+        "title": "View Template Typecheck Block",
+        "category": "Angular"
       }
     ],
+    "menus": {
+      "editor/context": [
+        {
+          "when": "resourceLangId == html || resourceLangId == ts",
+          "command": "angular.getTemplateTcb",
+          "group": "3_compare"
```
atscott, line 42: One more thing we might want to consider: we can define our own group here if we wanted all Angular commands to appear together in the context menu.

client/src/commands/getTcb.ts
```diff
@@ -0,0 +1,120 @@
...
+import * as vscode from 'vscode';
+
+import * as lspext from '../../common/lsp_extensions';
+import {AngularLanguageClient} from '../client';
+
+const TCB_HIGHLIGHT_DECORATION = vscode.window.createTextEditorDecorationType({
+  backgroundColor: new vscode.ThemeColor('editor.selectionHighlightBackground'),
+});
+
+class SourceMap {
```
kyliau, line 18: I don't think we need this class. Can't we just convert `lsp.Range` to `vscode.Range` by calling the `protocol2CodeConverter`?

kyliau (changes requested):

package.json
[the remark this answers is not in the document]
ayazhafiz: There is a nice listing at https://code.visualstudio.com/api/references/contribution-points#Sorting-of-groups.

server/yarn.lock
```diff
@@ -0,0 +1,38 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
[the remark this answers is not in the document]
ayazhafiz: Hm I don't remember why I had to do this.. will investigate more and follow up.

client/src/commands/getTcb.ts
```diff
@@ -0,0 +1,120 @@
...
+  // for the `ng` scheme next time `getTcb` is requested.
+  disposeNgProvider.dispose();
+}
+
+/**
+ * Command getTemplateTcb displays a typecheck block for the template a user has
+ * an active selection over, if any.
+ */
+export function getTemplateTcb(ngClient: AngularLanguageClient) {
+  const didRequestTcb = new vscode.EventEmitter<vscode.Uri>();
```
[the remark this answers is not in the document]
ayazhafiz: The return value of `getTemplateTcb` (namely `executeTemplateTcb`) is registered as a callback on the `getTcb` command. When that command is invoked `executeTemplateTcb` is called and we instantiate a document provider, alongside content for the TCB document.

However, the problem at this point is that VSCode updates documents lazily, and will not attempt to recompute a document's content without a notification that it needs to do so. So if we have already requested a TCB, say `a.typecheck.ts`, and then change something in our template and execute the `getTcb` command again, even though we will have updated the document provider and its content, the VSCode engine will not know to pick up these changes. Hence we associate an event emitter with each fresh document provider; VSCode will subscribe to the emitter, and when a new request for `getTcb` is made we fire the event to instruct VSCode to update the document content.

We could listen to changes in the active document, but that is more work than I want to include in this PR; it is a straightforward addition, but I think for these commits we should just get the command working without extra niceties.

Actually I just thought of a better model; I don't think there is any reason we have to reinstantiate a fresh document provider for each request. Again, I'll follow up after some more digging. Sorry, I have been very busy with other commitments so my responses here may be slow.

client/src/commands/getTcb.ts
ayazhafiz: Yes, but the problem is that we don't have `lsp.Range`. We have `ts.TextSpan`, which is a pair of character offsets (start, end) which we need to convert into line/column pairs.

common/lsp_extensions.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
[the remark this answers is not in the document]
ayazhafiz: Sound good 👍

client/src/commands/getTcb.ts
```diff
@@ -0,0 +1,120 @@
...
+    textDocument: client.code2ProtocolConverter.asTextDocumentIdentifier(activeEditor.document),
+    position: activeEditor.selection.active,
+  });
+  if (tcb === undefined) {
+    return;
+  }
+
+  // Allocate a provider of documents corresponding to the `ng` URI scheme,
+  // which we will use to provide a virtual document with the TCB contents.
+  const tcbProvider: vscode.TextDocumentContentProvider = {
```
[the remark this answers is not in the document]
ayazhafiz: That is one reason; another was to support syntax highlighting IIRC (but maybe Im wrong). I will follow up and add a comment.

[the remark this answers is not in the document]
ayazhafiz: Nevermind I think I did this just because of the readonly mode as you mentioned, not sure why I was thinking of highlighting. added a comment a link to the issue.

client/src/commands/getTcb.ts
atscott: I'm doing a similar conversion in my work with "find components for template" and I'm now using `tsTextSpanToLspRange` in `session.ts` and then from there, converting it to `vscode.Range` in the command. I think that would do the trick here as well?

client/src/commands/getTcb.ts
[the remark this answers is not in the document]
ayazhafiz: Okay, so the other option I was thinking is that we can have a single document provider and recompute the TCB closed over that provider (and filename, highlights). So it would look something like

```
function makeExecuteGetTcb(...) {
  let currentHighlights;
  let currentTcbFileName;
  let currentTcbContents;
  function update() {
      tcb = getTcb(fileName);
      currentHighlights = getHighlights(tcp);
      currentTcbFileName = tcb.fileName;
      currentTcbContents = tcb.content;
  }
  class TcbProvider extends vscode.TextDocumentProvider {
    onDidChange: new EventEmitter(),
    provideTextDocumentContent(fileName) {
      return currentTcbContents;
    }
  }
  const provider = new TcbProvider();
  vscode.register(provider);
  return function execute() {
    update();
    provider.onDidChange.fire(currentTcbFileName);
    const doc = vscode.open(currentTcbFileName);
    vscode.addDecorations(doc, currentHighlights);
  }
}
```

I'll let you decide which you like more; in either case it's not the most elegant because the filename that we have to compute is not known before we make the getTcb request.

Let me know if you have any more questions; I hope the last comment was able to clear up some confusion if any.

client/src/commands/getTcb.ts
ayazhafiz: We also cannot convert from TextSpan to Range on the server side because to do so we would have to assume that the TCB belongs to a script info; while this is true (IIRC), I don't think that's an assumption that the server should make with regard to how the language service handles TCBs and whether it does/doesn't attach them onto a project.

server/yarn.lock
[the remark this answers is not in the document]
ayazhafiz: Fixed

client/src/commands/getTcb.ts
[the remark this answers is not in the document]
ivanwonder: this is my thought with a single document provider. the only thing for the command `angular.getTemplateTcb` to do is just open the document by Uri. The provider will listen to the HTML change and refresh the tcb document.

```ts
class TcbDocumentContentProvider implements vscode.TextDocumentContentProvider, vscode.Disposable {
  private openFileSet = new Set<string>();

  constructor(private client: lsp.LanguageClient) {
    vscode.workspace.onDidOpenTextDocument(doc => {
      if (doc.uri.scheme === 'ngTcb') {
        this.openFileSet.add(doc.uri.query);
      }
    });
    vscode.workspace.onDidChangeTextDocument(doc => {
      if (this.openFileSet.has(doc.document.uri.path)) {
        const uri = vscode.Uri.parse('ngTcb: ngcheck.ts').with({query: doc.document.uri.path});
        this.onDidChangeEmitter.fire(uri);
      }
    });
    vscode.workspace.onDidCloseTextDocument(doc => {
      if (doc.uri.scheme === 'ngTcb') {
        this.openFileSet.delete(doc.uri.query);
      }
    });
    vscode.workspace.registerTextDocumentContentProvider('ngTcb', this);
  }

  dispose() {
    this.onDidChangeEmitter.dispose();
  }

  onDidChangeEmitter = new vscode.EventEmitter<vscode.Uri>();
  onDidChange = this.onDidChangeEmitter.event;

  provideTextDocumentContent(uri: vscode.Uri, token: vscode.CancellationToken):
      vscode.ProviderResult<string> {
    return this.client
        .sendRequest<GetTcbResponse>('ngTcb', {
          textDocument: lsp.TextDocumentIdentifier.create(uri.query),
          position: 0,
        })
        .then(tcb => tcb?.content);
  }
}
```
```ts
this.context.subscriptions.push(new TcbDocumentContentProvider(this.client));
```
```ts
function makeExecuteGetTcb(){
   let uri = vscode.Uri.parse('ngTcb: ngcheck.ts').with({query: activeEditor.document.uri.path});
    let doc = await vscode.workspace.openTextDocument(uri);  // calls back into the provider
    await vscode.window.showTextDocument(doc, {preview: false});
}
```

client/src/commands/getTcb.ts
kyliau: > we would have to assume that the TCB belongs to a script info ...

This will always be the case. Since the server owns the `ProjectService`, which in turn manages all `ScriptInfo`s, converting TS spans to LSP Range should be done by the server. We do the conversion for other methods too, like definition, references, etc.
In fact, it is a leak of abstraction to send TS spans down to the client, because the client should not be aware of TS at all. It only deals with LSP.

client/src/commands/getTcb.ts
kyliau: yeah, we don't have client-side tests / can't run them on CI.
I think a manual test here is sufficient, since this feature is not user-facing.

client/src/commands/getTcb.ts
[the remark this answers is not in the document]
kyliau: @ayazhafiz with your implementation in the comment above, does that mean we can only keep track of one tcb at a time? Can a provider manage multiple files at a time? I don't think so, but looking at @ivanwonder's suggestion it seems possible?

package.json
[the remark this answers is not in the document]
kyliau: +1 to Angular's own group.

package.json
```diff
@@ -133,7 +138,9 @@
     "test:syntaxes": "yarn compile:syntaxes-test && yarn build:syntaxes && jasmine dist/syntaxes/test/driver.js"
   },
   "dependencies": {
-    "@angular/language-service": "11.1.1",
+    "@angular/compiler": "11.2.0-next.0",
+    "@angular/compiler-cli": "11.2.0-next.0",
```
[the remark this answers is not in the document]
kyliau: With your change to `@angular/language-service`, we can delete these deps now.

server/package.json
```diff
@@ -15,7 +15,9 @@
     "ngserver": "./bin/ngserver"
   },
   "dependencies": {
-    "@angular/language-service": "11.1.1",
+    "@angular/compiler": "11.2.0-next.0",
```
kyliau, line 18: same comment as above, please delete.

kyliau (commented):
Thanks for looking into this and providing multiple suggestions.
I think it's pretty much good to go, once a few loose ends are taken care of:
1. Span conversion from TS to LSP should be done on the server
2. Create document provider once and use it throughout the lifetime of the extension.
3. Verify if the provider can manage multiple tcbs at a time.

As always, thank you for being so patient :)

client/src/commands/getTcb.ts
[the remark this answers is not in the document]
ayazhafiz: No, you can keep track of as many TCBs as you want if you have a single provider. In the example I give it only keeps track of one, but yes you can certainly keep track of several. However, I think "keeping track" of previously opened TCB documents is unnecessary, since (1) anytime someone requests a TCB you almost certainly want to make a request to the language server to get fresh TCB contents (I don't think trying to implement any kind of cache is worth it), and (2) if someone closes a TCB and then wants to reopen it, VSCode will have already cached the document and will keep that document content until the provider fires an event marking that the document should be reloaded.

So in my opinion, the only "important" thing is making sure that the event emitter for changed TCBs is properly exposed. To this end I feel the ways to  implement the provider here are the options between
- Creating a fresh document provider each time "getTcb" is requested, passing the same event emitter for requested file names (namely the file name corresponding to the requested TCB). This the current implementation.
- Making one document provider and closing over it, as done in Ivan's example or mine in the comment above Ivan's.

IMO, keeping track of previous document contents is unnecessary as explained above.

client/src/commands/getTcb.ts
ayazhafiz: Would it not be a leaky abstraction to assume that the typecheck block file is attached to the project? This is an internal detail of how the language service creates TCB files and may change later.

I agree the conversion boundary should happen on the server side, you bring up a great point.

ayazhafiz:
> 2. Create document provider once and use it throughout the lifetime of the extension.
> 3. Verify if the provider can manage multiple tcbs at a time.

Please see my comment above in your/Ivan/I's thread about this topic. I think the idea of managing multiple tabs is a misnomer, but creating a single document provider is trivial.

In any case, I do not have time to work on this this week. Feel free to take over and clean up the PR. Otherwise I can take a look this weekend.

kyliau:
Agree that the provider should manage only one tcb at a time, I think this is the behavior of the "markdown preview provider" in vscode. And we should only create a single provider.
Thanks for the great work @ayazhafiz and thanks @ivanwonder for the suggestion!
I'll clean this up and get it merged before the 11.2.0 release tomorrow.

client/src/commands/getTcb.ts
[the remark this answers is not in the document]
ivanwonder: > (1) anytime someone requests a TCB you almost certainly want to make a request to the language server to get fresh TCB contents (I don't think trying to implement any kind of cache is worth it),

I think I just record the HTML files for the TCB content and refresh the TCB content when the HTML files are changed. The TCB is managed by the uri, one TCB will be opened for the same uri. I have not tested it, but I think vscode will handle the uri. For example, when the user tries to view the TCB of file `files:/app/app.component.html`, we only need to open uri `ngtcb:/app/app.component.html`. The provider will handle the uri and provide the TCB content by sending the request to the service. and when the HTML changed, I will tell the provider the `ngtcb:/app/app.component.html` needs to be refreshed.

> (2) if someone closes a TCB and then wants to reopen it, VSCode will have already cached the document and will keep that document content until the provider fires an event marking that the document should be reloaded.

In my example, I think this will send a request to the service, because the vscode doesn’t know if the content is fresh, but it is ok to send a request(the language service has a cache for it) or add a cache here.

If I missed some info or misunderstood you, sorry for that. Because I think we can’t avoid listening to the HTML files, maybe you do it every time when open the TCB content. Just like you said it’s an option to achieve the function.

Some situations we need to be considered in the future.

- file deleted or renamed. If we need to close the TCB content or not.
- Sync scroll. Maybe we need to send a request to the service and scroll the TCB.
- file persistence. Maybe it’s not so important. After reload the vscode, the vscode will restore the layout and the opened file except for the TCB content.

So perhaps managed by uri is easier to understand for me.

client/src/providers.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+export class TcbContentProvider implements vscode.TextDocumentContentProvider {
+  /**
+   * Event emitter used to notify VSCode of a change to the TCB virtual document,
+   * prompting it to re-evaluate the document content. This is needed to bust
+   * VSCode's document cache if someone requests a TCB that was previously opened.
+   * https://code.visualstudio.com/api/extension-guides/virtual-documents#update-virtual-documents
+   */
+  private readonly onDidChangeEmitter = new vscode.EventEmitter<vscode.Uri>();
+  private tcfUri: vscode.Uri|null = null;
```
atscott, line 28: is this supposed to be `tcbUri` and `tcbContent` or am I not understanding what `tcf` stands for?

atscott (approved):
👍  LGTM

client/src/providers.ts
kyliau: renamed to `tcbFile` and `tcbContent`.

kyliau:
I cleaned this up a bit:
- create a new group called `angular` for the editor context menu
- make the separation between TS / LSP / vscode clearer
  - the server works with TS, but always sends the response to client in LSP
  - the client works with vscode, and is responsible for serializing / deserializing LSP
  - the common code should only be aware of LSP types
  - as a result, there are now three `GetTcbResponse` types - TS, LSP, and vscode
  - this allowed us to remove special logic for "source maps", the converter correctly handles all conversions
- create a single text document content provider for `ng` files, and use it for all requests.
- register the "get tcb" command as a text editor command (instead of a generic command), so that it is only available if there's an active editor

kyliau (approved):

PR #1016, opened 2020-12-10
fix: move ngcc operation to server

kyliau:
This commit fixes a bug in which ngcc could not be resolved from the client
when the extension is running in the production vscode environment.

This is because internally, the client depends on `require.resolve` to look
for `ngcc` given the directory where the `tsconfig.json` is.
However, vscode provides its own [`require` mechanism][1], which is different
from Node.js. In particular, the bug is due to `require.resolve` not honoring
the `paths` parameter provided to the method. See Node.js [documentation][2] on
`require.resolve`.

We did not catch this bug because the custom `require` is not used
in the extension development host.

In order to mitigate this, we have a few options:

1. Use an external [`resolve` package][3] in the client to perform node module resolution
2. Bundle ngcc with the client
3. Move ngcc operation to the server

(1) is not ideal because it adds an external dependency to the extension

(2) could potentially lead to compatibility issues due to different version
of ngcc in the extension vs user's project.

(3) would not only solve this issue, but also provides the additional benefit
of reducing the work needed to integrate the language server for clients
other than vscode (vim, emacs, for example).
This is because the client needs to do some extra work to find ngcc then
launch it, and report back to the server that the operation has completed.
With this change, the server handles ngcc completely, and no work is needed
on the client side.

As an aside, the initial motivation for implementing ngcc on the client
side is to provide the guarantee that the server is idempotent, i.e. it should
not produce the side effect of modifying files on disk, which ngcc does.
However, at this point, the benefits outweigh the cons, so ngcc operation is
moved to the server.

[1]: https://github.com/microsoft/vscode/blob/92adef0bac1260a6273443a2132428fdfc28a31f/src/vs/loader.js#L743-L745
[2]: https://nodejs.org/docs/latest-v12.x/api/modules.html#modules_require_resolve_request_options
[3]: https://www.npmjs.com/package/resolve

### DIFF
--- a/common/progress.ts
+++ b/common/progress.ts
@@ -1,0 +1,21 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {ProgressType} from 'vscode-jsonrpc';
+
+export type NgccProgress = {
+  done: false,
+  configFilePath: string,
+  message: string,
+}|{
+  done: true,
+  configFilePath: string,
+  success: boolean,
+};
+export const NgccProgressToken = 'ngcc';
+export const NgccProgressType = new ProgressType<NgccProgress>();
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -10,10 +10,12 @@ import * as ts from 'typescript/lib/tsserverlibrary';
 import * as lsp from 'vscode-languageserver';
 
 import {ServerOptions} from '../common/initialize';
-import * as notification from '../common/notifications';
+import {ProjectLanguageService, ProjectLoadingFinish, ProjectLoadingStart} from '../common/notifications';
+import {NgccProgressToken, NgccProgressType} from '../common/progress';
 
 import {tsCompletionEntryToLspCompletionItem} from './completion';
 import {tsDiagnosticToLspDiagnostic} from './diagnostic';
+import {resolveAndRunNgcc} from './ngcc';
 import {ServerHost} from './server_host';
 import {filePathToUri, isConfiguredProject, lspPositionToTsPosition, lspRangeToTsPositions, tsTextSpanToLspRange, uriToFilePath} from './utils';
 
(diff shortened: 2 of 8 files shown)


ayazhafiz:
This would mean that the server process must always run under the same filesystem as the client. This is a reasonable assumption for the vast majority of cases folks are using the language server today but may become a blocker for other use cases later on (as remote process in google3, in an online IDE).

kyliau:
> This would mean that the server process must always run under the same filesystem as the client. This is a reasonable assumption for the vast majority of cases folks are using the language server today but may become a blocker for other use cases later on (as remote process in google3, in an online IDE).

@ayazhafiz Yeah, you're right. The good thing is google3 doesn't need ngcc because all dependencies are compiled from source. Also there's a project in the works called `Ng Linker`. When it is done, we don't need `ngcc` anymore (in about a year's time probably).

server/src/session.ts
```diff
@@ -600,9 +632,8 @@ export class Session {
     if (this.ivy && isConfiguredProject(project)) {
       // Keep language service disabled until ngcc is completed.
       project.disableLanguageService();
-      this.connection.sendNotification(notification.RunNgcc, {
-        configFilePath: project.getConfigFilePath(),
-      });
+      // Do not wait on this promise otherwise we'll be blocking other requests
+      this.runNgcc(project.getConfigFilePath());
```
ayazhafiz, line 636: Do you think it would be good to add an [unhandled promise rejection handler](https://nodejs.org/api/process.html#process_event_unhandledrejection)?

ayazhafiz (approved):

common/progress.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+import {ProgressType} from 'vscode-jsonrpc';
+
+export type NgccProgress = {
+  done: false; configFilePath: string; message: string;
+}|{
+  done: true;
+  configFilePath: string;
+};
```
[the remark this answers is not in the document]
atscott: @kyliau - `type` declarations should use commas instead of semicolons. I think this might fix the clang-format weirdness.
go/tsstyle#member-property-declarations

server/src/session.ts
```diff
@@ -111,18 +113,48 @@ export class Session {
...
       this.error(
           `Failed to run ngcc for ${configFilePath}:\n` +
-          `${params.error}\n` +
-          `Language service will remain disabled.`);
+          `    ${e.message}\n` +
+          `    Language service will remain disabled.`);
+    } finally {
+      this.connection.sendProgress(NgccProgressType, NgccProgressToken, {
+        done: true,
+        configFilePath,
```
atscott, line 146: Does this mean the client isn't really notified when an error happens? It would just show up in server logs and the client would see that ngcc is 'done'?

atscott (approved):

kyliau:
Show error message when ngcc fails. Also added an option to reveal the extension output.

Screen Shot 2020-12-11 at 10 28 19 AM

PR #1017, opened 2020-12-11
fix: restart the server with new options on configuration change

kyliau:
Currently, when users change their vscode configuration (say when they enable
Ivy mode), they'll have to manually reload the window to force the new
configuration to take effect.

However, we cannot just start and stop the `LanguageClient` connection
because the server options passed to the client on startup is immutable.

To fix this, we wrap the `LanguageClient` in our own class
`AngularLanguageClient`, and reconstruct the server options every time the
server is restarted.

Code in `client/src/extension.ts` is copied to `client/src/client.ts` to avoid circular dependency between `extension.ts` and `commands.ts`.

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -1,0 +1,239 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import * as lsp from 'vscode-languageclient';
+
+import {ProjectLoadingFinish, ProjectLoadingStart} from '../common/notifications';
+import {NgccProgress, NgccProgressToken, NgccProgressType} from '../common/progress';
+
+import {ProgressReporter} from './progress-reporter';
+
+export class AngularLanguageClient implements vscode.Disposable {
+  private client: lsp.LanguageClient|null = null;
+  private readonly disposables: vscode.Disposable[] = [];
+  private readonly outputChannel: vscode.OutputChannel;
+  private readonly clientOptions: lsp.LanguageClientOptions;
+  private readonly name = 'Angular Language Service';
+
+  constructor(private readonly context: vscode.ExtensionContext) {
+    this.outputChannel = vscode.window.createOutputChannel(this.name);
+    // Options to control the language client
+    this.clientOptions = {
+      // Register the server for Angular templates and TypeScript documents
+      documentSelector: [
+        // scheme: 'file' means listen to changes to files on disk only
+        // other option is 'untitled', for buffer in the editor (like a new doc)
+        {scheme: 'file', language: 'html'},
+        {scheme: 'file', language: 'typescript'},
+      ],
+      synchronize: {
+        fileEvents: [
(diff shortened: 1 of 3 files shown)


client/src/client.ts
```diff
@@ -0,0 +1,235 @@
...
+    );
+    this.disposables.push(this.client.start());
+    await this.client.onReady();
+    // Must wait for the client to be ready before registering notification
+    // handlers.
+    registerNotificationHandlers(this.client);
+  }
+
+  /**
+   * Kill the language server and perform some clean ups.
```
ayazhafiz, line 80:
```suggestion
   * Kill the language client and perform some clean ups.
```

client/src/client.ts
```diff
@@ -0,0 +1,235 @@
...
+      client.sendNotification(NgccComplete, {
+        configFilePath,
+        success: false,
+        error: e.message,
+      });
+    }
+  });
+}
+
+/**
```
ayazhafiz, line 145: Assume the rest of the file is just copied over from existing impl?

client/src/commands.ts
```diff
@@ -23,30 +23,39 @@ interface Command {
...
-        // TODO: We could show a MessageItem to help users automatically update
-        // the configuration option then restart the server, but we currently do
-        // not reload the server options when restarting the server.
-        vscode.window.showErrorMessage(
-            `Angular Server logging is off. Please set 'angular.log' and reload the window.`);
+        // Show a MessageItem to help users automatically update the
+        // configuration option then restart the server.
+        const selection = await vscode.window.showErrorMessage(
+            `Angular server logging is off. Please set 'angular.log' and restart the server.`,
+            'Enable logging and restart Angular server',
```
ayazhafiz, line 49:
```suggestion
        const selection = await vscode.window.showErrorMessage(
            `Angular server logging is off. Please set 'angular.log' and restart the server.`,
            'Enable verbose logging and restart Angular server',
            `Set 'angular.log' manually (server will not be restarted)`,
```

client/src/extension.ts
```diff
@@ -6,193 +6,24 @@
...
-      client.sendNotification(notification.NgccComplete, {
-        configFilePath,
-        success: false,
-        error: e.message,
-      });
-    }
+  // Restart the server on configuration change.
+  const disposable = vscode.workspace.onDidChangeConfiguration(async () => {
+    await client.stop();
+    await client.start();
```
ayazhafiz, line 24: Is it worth displaying a message that the server is restarting? I think no, what do you think?

ayazhafiz (approved):

client/src/client.ts
kyliau: yes, no change here.

kyliau: Code in `client/src/extension.ts` is copied to `client/src/client.ts` to avoid circular dependency between `extension.ts` and `commands.ts`.

client/src/extension.ts
kyliau: yeah, I don't think we need to display a message, because when the server restarts, it'll actually show "Initializing Angular language features ...". I think it's a good enough signal to indicate the server has restarted.

client/src/commands.ts
kyliau: Actually, we don't need to add the second option, because if users want to set the verbosity manually, they could just ignore this message or hit `escape`.

atscott (approved):

PR #1058, opened 2021-01-12
fix: prevent project from closing when only a template file is open

atscott:
After opening a file, TypeScript does a cleanup step to remove/close
"orphaned" projects: https://github.com/microsoft/TypeScript/blob/a894f8ad2b0ff35b1bcfd91defba994136c68b18/src/server/editorServices.ts#L3337
As part of this, TS checks to make sure there are no open files for the
configured project: https://github.com/microsoft/TypeScript/blob/a894f8ad2b0ff35b1bcfd91defba994136c68b18/src/server/project.ts#L2333
If the only open file for the project is an HTML file, TS will not know
about it because we do not register those with the project (instead, we
associate the template file's `ScriptInfo` with the project, but do not
make the association in the other direction).
Because we do not have this two way association between the project and
the HTML `ScriptInfo`s, TypeScript closes the configured project on
actions like "go to definition" that open a `d.ts` file outside the
project. This is problematic for two particular reasons:
1. The project should not be closed in the first place. We will take a
performance hit when the user inevitably goes back to the template file
because now the project has to be re-opened.
2. The template file's `ScriptInfo` still tracks the closed project in
its `containingProjects` array. As part of the cleanup for closing a
project, TS would detach the `ScriptInfo`s from the project
(https://github.com/microsoft/TypeScript/blob/a894f8ad2b0ff35b1bcfd91defba994136c68b18/src/server/project.ts#L752)
but again, because we don't have the two-way association as mentioned
above, this does not happen. This results in a crash the next time a
request is made from the template file.

This commit addresses both of the issues above.

First, we prevent the project from being closed by creating an external project associated
with the configured project for the template file. When
TypeScript checks if a project has open refs/files, it returns `true` if
there are external project references
(https://github.com/microsoft/TypeScript/blob/a894f8ad2b0ff35b1bcfd91defba994136c68b18/src/server/project.ts#L2334).
This means that the clean up after opening a new file process as
mentioned above will not result in our project for the template being
closed.

Second, when retrieving a script info for a file we check if the project
as been closed. If it has, we detach the script info from the project,
log an error, and simply return `undefined`. To our knowledge, this
should not happen because of the first solution we implemented, but it
would prevent future crashes if there are other scenarios we missed.

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -44,6 +44,7 @@ export class Session {
   private readonly projectService: ts.server.ProjectService;
   private readonly logger: ts.server.Logger;
   private readonly ivy: boolean;
+  private readonly configuredProjToExternalProj = new Map<string, string>();
   private diagnosticsTimeout: NodeJS.Timeout|null = null;
   private isProjectLoading = false;
 
@@ -307,6 +308,7 @@ export class Session {
       scriptInfo.detachAllProjects();
       scriptInfo.attachToProject(project);
     }
+    this.createExternalProject(project);
 
     return project;
   }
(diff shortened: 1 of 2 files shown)


server/src/session.ts
```diff
@@ -377,13 +379,61 @@ export class Session {
...
     if (!filePath) {
       return;
     }
     this.projectService.closeClientFile(filePath);
+    this.maybeCloseExternalProject(filePath);
+  }
+
+  /**
+   * We open external projects for files so that we can prevent TypeScript from closing a project
+   * when it shouldn't be. This function checks if there are no longer any open files in the project
```
kyliau, line 411: ... prevent TypeScript from closing a project when there is open external HTML template that still references the project.

kyliau (approved):

PR #1065, opened 2021-01-15
feat: Add renaming capability

atscott:
This commit adds rename capability to the Angular Language Service
extension. It's important to note that vscode only uses results from a
single rename provider. We cannot currently ensure that Angular takes
precedence over TypeScript when renaming from TS files
(see microsoft/vscode#115354) so this feature
is only available in _external_ template files (it is neither available for inline
templates nor other locations in TS files).

### DIFF
--- a/integration/lsp/test_utils.ts
+++ b/integration/lsp/test_utils.ts
@@ -17,6 +17,7 @@ const PACKAGE_ROOT = resolve(__dirname, '../../..');
 const PROJECT_PATH = `${PACKAGE_ROOT}/integration/project`;
 export const APP_COMPONENT = `${PROJECT_PATH}/app/app.component.ts`;
 export const FOO_TEMPLATE = `${PROJECT_PATH}/app/foo.component.html`;
+export const FOO_COMPONENT = `${PROJECT_PATH}/app/foo.component.ts`;
 
 export interface ServerOptions {
   ivy: boolean;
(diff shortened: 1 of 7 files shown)


kyliau (approved):

jesuspa17:
:partying_face: :+1:

atscott:
Adding the blocked label: It turns out that the dependency on the built in extension doesn't work as desired. If the first file opened is an `html` file, then Angular registers as a rename provider. Then, when a typescript file is open, the built in extension registers as a provider. This means that the built-in will at that point win priority for renames from TS files.

atscott:
Tracking vscode issue for rename provider configuration: https://github.com/microsoft/vscode/issues/115354

integration/lsp/test_utils.ts
```diff
@@ -70,11 +71,12 @@ export function initializeServer(client: MessageConnection): Promise<lsp.Initial
   });
 }
 
-export function openTextDocument(client: MessageConnection, filePath: string) {
+export function openTextDocument(
+    client: MessageConnection, filePath: string, languageId = 'typescript') {
```
kyliau, line 75: nit: I think it's better we just infer languageId from filePath extension.

kyliau (approved):

integration/lsp/test_utils.ts
atscott: 👍 Yea, that's much better. Thanks for the suggestion

PR #1097, opened 2021-02-03
Suggest strict templates

### LINKED ISSUE #1053: Prompt users to enable strictTemplates
Ivy language service works best when `fullTemplateTypeCheck` is enabled. This flag could be provided in `angularCompilerOptions` in `tsconfig.json`.

If the flag is not on, some features (go to definition, hover, etc) might not work in certain situations, because the necessary type information is not present in the typecheck files.

This could be confusing or even misleading for users because they would think the language service is broken.

When this flag is not turned on, we need to tell users that language service is not working at its full capability.

There are a few ways we could do this:
1. Show a pop-up window in the editor at startup
2. Show a diagnostic message when user tries to perform an operation but get no results.
3. Show a hint in user's code.

## Show a pop-up window in the editor

Screen Shot 2021-01-07 at 1 26 52 PM

Pro: easiest to implement
Con: could be a little distracting since this is shown every time the project is loaded

Necessary work:
1. Expose the Angular compiler options to the language server
2. Language server sends a notification to the client
3. Client shows the message in a pop up

## Show a diagnostic message when user tries to perform an operation

Screen Shot 2021-01-07 at 1 34 18 PM

Show an `Information` diagnostic when users try to perform an operation but get no results.
This diagnostic will show up in the "Problems" panel.

Pro: Less distracting, because diagnostic is only shown when user tries to do something.
Pro: Does not involve client / server, this is purely on the language service side.
Con: Not straightforward to implement

## Show a hint in user's code

This is similar to the diagnostic approach, except the message does not show up in the "Problems" panel.
The message shows up when user hovers over the `...` in the editor.

Screen Shot 2021-01-07 at 1 34 54 PM

Pro: less distracting
Con: Not straightforward to implement

# Bonus
- Add a "quick fix" to automatically update `tsconfig.json` for users.

# Suggestion
I suggest we do (1) since the impact/work ratio is the highest. We could implement (2) or (3) once the core features are stable.

kyliau:
This commit adds a prompt to encourage users turn on `strictTemplates` so
that they can get the most out of Angular language service.

Screen Shot 2021-02-03 at 11 06 10 AM

PR close #1053

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -11,7 +11,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import * as lsp from 'vscode-languageclient/node';
 
-import {ProjectLoadingFinish, ProjectLoadingStart} from '../common/notifications';
+import {ProjectLoadingFinish, ProjectLoadingStart, SuggestStrictMode, SuggestStrictModeParams} from '../common/notifications';
 import {NgccProgress, NgccProgressToken, NgccProgressType} from '../common/progress';
 
 import {ProgressReporter} from './progress-reporter';
@@ -76,7 +76,7 @@ export class AngularLanguageClient implements vscode.Disposable {
     await this.client.onReady();
     // Must wait for the client to be ready before registering notification
     // handlers.
-    registerNotificationHandlers(this.client);
+    registerNotificationHandlers(this.client, this.context);
     registerProgressHandlers(this.client, this.context);
   }
 
@@ -104,8 +104,9 @@ export class AngularLanguageClient implements vscode.Disposable {
   }
 }
 
-function registerNotificationHandlers(client: lsp.LanguageClient) {
-  client.onNotification(ProjectLoadingStart, () => {
+function registerNotificationHandlers(
+    client: lsp.LanguageClient, context: vscode.ExtensionContext) {
+  const disposable1 = client.onNotification(ProjectLoadingStart, () => {
     vscode.window.withProgress(
         {
           location: vscode.ProgressLocation.Window,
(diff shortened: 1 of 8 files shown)


client/src/client.ts
```diff
@@ -104,8 +104,9 @@ export class AngularLanguageClient implements vscode.Disposable {
   }
 }
 
-function registerNotificationHandlers(client: lsp.LanguageClient) {
-  client.onNotification(ProjectLoadingStart, () => {
+function registerNotificationHandlers(
+    client: lsp.LanguageClient, context: vscode.ExtensionContext) {
+  const d1 = client.onNotification(ProjectLoadingStart, () => {
```
atscott, line 109: nit: d1 and d2 aren't great good variable names

client/src/client.ts
```diff
@@ -116,6 +117,19 @@ function registerNotificationHandlers(client: lsp.LanguageClient) {
         }),
     );
   });
+
+  const d2 = client.onNotification(SuggestStrictMode, async (params: SuggestStrictModeParams) => {
+    const selection = await vscode.window.showInformationMessage(
+        params.message,
+        'Open tsconfig.json',
+    );
+    if (selection) {
```
atscott, line 126:
```suggestion
    const openTsConfig = 'Open tsconfig.json';
    const selection = await vscode.window.showInformationMessage(
        params.message,
        openTsConfig,
    );
    if (selection === openTsConfig) {
```
While I know the API for `showInformationMessage`, I think this might be a bit more clear. WDYT?

server/src/session.ts
```diff
@@ -201,6 +201,23 @@ export class Session {
     project.getLanguageService().getSemanticDiagnostics(fileName);
   }
 
+  private suggestStrictMode(project: ts.server.Project) {
```
atscott, line 204:
```suggestion
  private handleCompilerOptionsDiagnostics(project: ts.server.Project) {
```
What do you think of making this function name a bit more generic?

atscott (approved):
👍  Nice

client/src/client.ts
kyliau: renamed to `disposable1` and `disposable2`.

atscott (approved):

PR #1099, opened 2021-02-03
feat: Prompt to use the Ivy Language Service if VE is detected

### LINKED ISSUE #1095: Add notification to prompt users to enable Ivy experimental LS
As an extra nudge to opt in to the new language service, provide a prompt for users to enable it.

atscott:
In order to drive adoption of the new Ivy Language Service before it
becomes the default in v12, this commit adds a prompt for users to
enable it if they are using `@angular/core` v9+.

Fixes #1095

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -11,7 +11,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import * as lsp from 'vscode-languageclient/node';
 
-import {ProjectLoadingFinish, ProjectLoadingStart, SuggestStrictMode, SuggestStrictModeParams} from '../common/notifications';
+import {ProjectLoadingFinish, ProjectLoadingStart, SuggestIvyLanguageService, SuggestIvyLanguageServiceParams, SuggestStrictMode, SuggestStrictModeParams} from '../common/notifications';
 import {NgccProgress, NgccProgressToken, NgccProgressType} from '../common/progress';
 
 import {ProgressReporter} from './progress-reporter';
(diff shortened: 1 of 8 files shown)


server/src/session.ts
```diff
@@ -830,6 +834,12 @@ export class Session {
     } else {
       // Immediately enable Legacy/ViewEngine language service
       this.info(`Enabling VE language service for ${projectName}.`);
+      if (!this.ivy && this.angularCoreVersion.major >= 9) {
```
kyliau, line 837: We should resolve the version of `@angular/core` here, since we already know that `@angular/core/core.d.ts` exists. This would allow us to run the check for each project.

client/src/client.ts
```diff
@@ -136,7 +136,26 @@ function registerNotificationHandlers(
...
+        }
+
+        const enableIvy = 'Enable';
+        const selection = await vscode.window.showInformationMessage(
+            params.message,
+            enableIvy,
+            'Ignore',
+        );
+        if (selection === enableIvy) {
+          config.update('angular.experimental-ivy', true);
```
kyliau, line 153: Does the language service automatically restart once the config is updated?

server/src/session.ts
```diff
@@ -830,6 +834,12 @@ export class Session {
...
+        this.connection.sendNotification(SuggestIvyLanguageService, {
+          message: 'An old version of the Angular Language Service is enabled.' +
```
kyliau, line 839: nit: `enabled` -> `detected`?
Or just omit the first sentence, and ask users 

> Would you like to enable the new Ivy-native language service to get the latest features and bug fixes?

We can link to the blog post once it's out.

kyliau (changes requested):

client/src/client.ts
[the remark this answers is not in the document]
atscott: The default is `WorkspaceFolder` - I've updated this to be global. We might want to discuss how we want to approach it. Maybe the `ignore` selection is global and the `enable` option is per workspace? WDYT?

atscott: Yep, it looks like it restarts whenever any configuration changes.

server/src/session.ts
```diff
@@ -850,6 +855,17 @@ export class Session {
     } else {
       // Immediately enable Legacy/ViewEngine language service
       this.info(`Enabling VE language service for ${projectName}.`);
+      if (!this.angularCoreVersion.has(project)) {
+        const angularCore = resolveAngularCore(this.ngProbeLocations);
```
kyliau, line 859: `ngProbeLocations` is a collection of paths which we would like to resolve `@angular/language-service` from, and it is not the same as the location of `@angular/core` for the project. For example, the probe locations include `ngdk` and the extension directory where `@angular/language-service` is bundled.
Here, we should resolve `@angular/core` from 
```suggestion 
        const angularCore = resolve(
          '@angular/core', 
          project.getFileNames().find(f => f.endsWith('@angular/core/core.d.ts')!
        );
```
The `@angular/core/core.d.ts` file is guaranteed to exist because it's checked here:
https://github.com/angular/vscode-ng-language-service/blob/0abe4094f4a9af45c53dcabb7e4cf470dedcc50e/server/src/session.ts#L864-L865
We should just refactor `checkIsAngularProject()` to return the file name if the project is Angular.
https://github.com/angular/vscode-ng-language-service/blob/0abe4094f4a9af45c53dcabb7e4cf470dedcc50e/server/src/session.ts#L856-L860

atscott: Done.

kyliau (approved):

PR #1100, opened 2021-02-03
fix: disable rename feature when strictTemplates is disabled

atscott:
When `strictTemplates` is disabled, we know that the compiler's view of a template
is not completely "accurate". That is, the compiler's generated Type
Check Blocks will not have sufficient type information to properly
perform a rename operation.

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -49,6 +49,13 @@ export class Session {
   private readonly logToConsole: boolean;
   private diagnosticsTimeout: NodeJS.Timeout|null = null;
   private isProjectLoading = false;
+  /**
+   * Tracks which `ts.server.Project`s have the renaming capability disabled.
+   *
+   * If we detect the compiler options diagnostic that suggests enabling strict mode, we want to
+   * disable renaming because we know that there are many cases where it will not work correctly.
+   */
+  private renameDisabledProjects: WeakSet<ts.server.Project> = new WeakSet();
 
   constructor(options: SessionOptions) {
     this.logger = options.logger;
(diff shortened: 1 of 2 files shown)


server/src/session.ts
```diff
@@ -49,6 +49,12 @@ export class Session {
   private readonly logToConsole: boolean;
   private diagnosticsTimeout: NodeJS.Timeout|null = null;
   private isProjectLoading = false;
+  /**
+   * By default, the rename feature is enabled. However, if we detect the compiler options
+   * diagnostic that suggests enabling strict mode, we want to disable renaming because we know that
+   * there are many cases where it will not work correctly.
+   */
+  private disableRenaming = false;
```
kyliau, line 57: This should be a per-project setting, so we probably want a `WeakMap<ts.server.Project, boolean>` here.

atscott: Good catch - I made this a `Set<ts.server.Project>` to just track which projects have renaming disabled.

server/src/session.ts
```diff
@@ -49,6 +49,13 @@ export class Session {
   private readonly logToConsole: boolean;
   private diagnosticsTimeout: NodeJS.Timeout|null = null;
   private isProjectLoading = false;
+  /**
+   * Tracks which `ts.server.Project`s have the renaming capability disabled.
+   *
+   * If we detect the compiler options diagnostic that suggests enabling strict mode, we want to
+   * disable renaming because we know that there are many cases where it will not work correctly.
+   */
+  private renameDisabledProjects: Set<ts.server.Project> = new Set();
```
kyliau, line 58: Should we use a `WeakSet` here so that projects that are closed can be gc-ed?

atscott: 👍  Done.

kyliau (approved):

PR #1107, opened 2021-02-12
feat: add command to go to component(s) from external template

atscott:
This commit adds a command to locate the components which reference a
given external template via their `templateUrl`s. If there is only one
component, immediately navigate to the component class declaration. If
there are multiple components, show them each in the same manner as
"find references" would.

### DIFF
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -127,27 +127,18 @@ function goToComponentWithTemplateFile(ngClient: AngularLanguageClient): Command
     id: 'angular.goToComponentWithTemplateFile',
     isTextEditorCommand: true,
     async execute(textEditor: vscode.TextEditor) {
-      const componentLocations = await ngClient.getComponentsForOpenExternalTemplate(textEditor);
-      if (componentLocations === undefined) {
+      const locations = await ngClient.getComponentsForOpenExternalTemplate(textEditor);
+      if (locations === undefined) {
         return;
       }
 
-      const locations: vscode.Location[] =
-          componentLocations.map(location => new vscode.Location(location.uri, location.range));
-      // If there is more than one component that references the template, show them all. Otherwise
-      // go to the component immediately.
-      if (locations.length > 1) {
-        vscode.commands.executeCommand(
-            'editor.action.showReferences',
-            textEditor.document.uri,
-            new vscode.Position(
-                textEditor.selection.start.line, textEditor.selection.start.character),
-            locations,
-        );
-      } else {
-        const document = await vscode.workspace.openTextDocument(locations[0].uri);
-        await vscode.window.showTextDocument(document, {selection: locations[0].range});
-      }
+      vscode.commands.executeCommand(
+          'editor.action.goToLocations',
+          textEditor.document.uri,
+          textEditor.selection.active,
+          locations,
+          'peek', /** what to do when there are multiple results */
+      );
     },
   };
 }
--- a/client/src/commands.ts
+++ b/client/src/commands.ts
@@ -117,6 +117,42 @@ function getTemplateTcb(
 }
 
 /**
+ * Command goToComponentWithTemplateFile finds components which reference an external template in
+ * their `templateUrl`s.
+ *
+ * @param ngClient LSP client for the active session
+ */
+function goToComponentWithTemplateFile(ngClient: AngularLanguageClient): Command {
+  return {
+    id: 'angular.goToComponentWithTemplateFile',
+    isTextEditorCommand: true,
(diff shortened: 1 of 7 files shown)


client/src/commands.ts
```diff
@@ -116,6 +116,42 @@ function getTemplateTcb(
...
+    id: 'angular.goToComponentWithTemplateFile',
+    isTextEditorCommand: true,
+    async execute(textEditor: vscode.TextEditor) {
+      const componentLocations = await ngClient.getComponentsForOpenExternalTemplate(textEditor);
+      if (componentLocations === undefined) {
+        return;
+      }
+
+      const locations: vscode.Location[] =
+          componentLocations.map(location => new vscode.Location(location.uri, location.range));
```
kyliau, line 136: Consider moving this operation to `getComponentsForOpenExternalTemplate`.

client/src/commands.ts
```diff
@@ -116,6 +116,42 @@ function getTemplateTcb(
...
+        vscode.commands.executeCommand(
+            'editor.action.showReferences',
+            textEditor.document.uri,
+            new vscode.Position(
+                textEditor.selection.start.line, textEditor.selection.start.character),
+            locations,
+        );
+      } else {
+        const document = await vscode.workspace.openTextDocument(locations[0].uri);
+        await vscode.window.showTextDocument(document, {selection: locations[0].range});
```
kyliau, line 149: Does this replace the current editor or show it in a separate panel?
Go to definition from `templateUrl` opens the file in a separate editor, so the behavior here should be consistent with that as well.

common/requests.ts
```diff
@@ -8,6 +8,17 @@
...
+export const GetComponentsWithTemplateFile = new lsp.RequestType<
+    GetComponentsWithTemplateFileParams, GetComponentsWithTemplateFileResponse,
+    /* error */ void>('angular/getComponentsWithTemplateFile');
+
+export interface GetComponentsWithTemplateFileParams {
+  textDocument: lsp.TextDocumentIdentifier;
+}
+
+/** An array of locations that represent component declarations. */
+export type GetComponentsWithTemplateFileResponse = Array<{uri: lsp.DocumentUri, range: lsp.Range}>;
```
kyliau, line 20: Consider reusing LSP's `Location` or `LocationLink`
```ts
/**
 * Represents a location inside a resource, such as a line
 * inside a text file.
 */
export interface Location {
    uri: DocumentUri;
    range: Range;
}
```

server/src/session.ts
```diff
@@ -160,6 +161,24 @@ export class Session {
...
+    if (lsInfo === undefined) {
+      return undefined;
+    }
+    const {languageService, scriptInfo} = lsInfo;
+    const documentSpans = languageService.getComponentLocationsForTemplate(scriptInfo.fileName);
+    const results: GetComponentsWithTemplateFileResponse = [];
+    for (const documentSpan of documentSpans) {
+      const scriptInfo = this.projectService.getScriptInfo(documentSpan.fileName);
+      const range =
+          scriptInfo ? tsTextSpanToLspRange(scriptInfo, documentSpan.textSpan) : EMPTY_RANGE;
```
kyliau, line 176: If the scriptInfo does not exist for the span then we should just skip this result.

kyliau (changes requested):

client/src/commands.ts
atscott: Updated to use the `goToLocations` instead, which handles multiple results already rather than having this logic here. also tested to ensure this operates the same as a "go to definition" operation with respect to pinned/unpinned files.

client/src/commands.ts
```diff
@@ -116,6 +116,34 @@ function getTemplateTcb(
...
+      const locations = await ngClient.getComponentsForOpenExternalTemplate(textEditor);
+      if (locations === undefined) {
+        return;
+      }
+
+      vscode.commands.executeCommand(
+          'editor.action.goToLocations',
+          textEditor.document.uri,
+          new vscode.Position(
+              textEditor.selection.start.line, textEditor.selection.start.character),
```
kyliau, line 139: Is this referring to the selection in the source document?
I think we can use `textEditor.selection.active`, which is already a `vscode.Position`.
```ts
    /**
     * Represents a text selection in an editor.
     */
    export class Selection extends Range {

        /**
         * The position at which the selection starts.
         * This position might be before or after [active](#Selection.active).
         */
        anchor: Position;

        /**
         * The position of the cursor.
         * This position might be before or after [anchor](#Selection.anchor).
         */
        active: Position;

        // ....
}
```

kyliau (approved):

PR #1118, opened 2021-02-16
fix: ensure project language service is the Angular LS

### LINKED ISSUE #1110: Ivy LS: duplicate results for autocomplete, quick info, etc outside of templates
**Describe the bug**

The latest version, v11.2.0, seems to have added 2 regression bugs:
1) One that prevents autocompletion of class variables in templates
2) Another which causes duplicate auto-complete options outside templates
I'm including both bug reports here as I feel they may be related.

**To Reproduce**

1. Upgrade to the latest version of Angular-CLI
2. Create a new project

Bug 1:
1. Go to app.component.html
2. Remove all HTML
3. attempt to get the language service to autocomplete options for class members (eg try to get it to detect `title`)
image

Bug 2:
1. Wait for up to 1 minute
2. Anywhere in the project, for example, the `app.module.ts`, press ctrl+space
3. Note that all auto-completed options have been duplicated
image

**Expected behavior**

1. I expect to get autocompletion in the template (as seen in bug 1)
2. I expect not to see duplicate autocomplete options outside of templates (as seen in bug 2)

**Logs**

Please attach two logs:

1. Console output (Go to toolbar --> View --> Output, and copy the contents)
- Nothing showing 
2. Log file (Path is printed to the console output at startup)
- Not sure what this means, sorry

**Additional context**

I'm using Windows 10
Angular CLI version 11.2.0
I tried to reproduce this in another window (in a React project) and could not do so
If I downgrade to 11.1.3 things all work as expected

atscott:
Add check to ensure the returned language service for the project is the Angular
Language Service. This removes the need for a type assertion and also prevents
the native TS LS from taking the place of ours if there is an issue with the
plugin registration (which would result in duplicate results in TS files).

Fixes #1110.

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -708,8 +708,12 @@ export class Session {
       this.logger.info(`Failed to get language service for closed project ${project.projectName}.`);
       return undefined;
     }
+    const languageService = project.getLanguageService();
+    if (!isNgLs(languageService)) {
+      return undefined;
+    }
     return {
-      languageService: project.getLanguageService() as NgLanguageService,
+      languageService,
       scriptInfo,
     };
   }
@@ -936,4 +940,8 @@ function toArray<T>(it: ts.Iterator<T>): T[] {
     results.push(itResult.value);
   }
   return results;
+}
+
+function isNgLs(ls: ts.LanguageService|NgLanguageService): ls is NgLanguageService {
+  return 'getTcb' in ls;
 }


server/src/session.ts
```diff
@@ -708,8 +708,12 @@ export class Session {
       this.logger.info(`Failed to get language service for closed project ${project.projectName}.`);
       return undefined;
     }
+    const languageService = project.getLanguageService();
+    if (!('getTcb' in languageService)) {
+      return undefined;
+    }
```
ivanwonder, line 714: nit: Looks like the ts infer `languageService` to `never`, why not use type guards?
```ts
  function isNgLs(ls: ts.LanguageService): ls is NgLanguageService {
    return 'getTcb' in ls;
  }
```

atscott: Good suggestion, thanks!

kyliau (approved):

PR #1161, opened 2021-02-26
fix: detect @angular/core in google3 and don't run ngcc

kyliau:
Refactors the detection logic for Angular project to take into account
`@angular/core` in google3.
If internal `@angular/core` is detected, do not run ngcc.

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -180,63 +180,23 @@ export class Session {
     return results;
   }
 
-  private async runNgcc(configFilePath: string) {
-    this.connection.sendProgress(NgccProgressType, NgccProgressToken, {
-      done: false,
-      configFilePath,
-      message: `Running ngcc for project ${configFilePath}`,
-    });
-
-    let success = false;
-
-    try {
-      await resolveAndRunNgcc(configFilePath, {
-        report: (msg: string) => {
-          this.connection.sendProgress(NgccProgressType, NgccProgressToken, {
-            done: false,
-            configFilePath,
-            message: msg,
-          });
-        },
-      });
-      success = true;
-    } catch (e) {
-      this.error(
-          `Failed to run ngcc for ${configFilePath}:\n` +
-          `    ${e.message}\n` +
-          `    Language service will remain disabled.`);
-    } finally {
-      this.connection.sendProgress(NgccProgressType, NgccProgressToken, {
-        done: true,
-        configFilePath,
-        success,
-      });
+  private enableLanguageServiceForProject(project: ts.server.Project, angularCore: string) {
+    const {projectName} = project;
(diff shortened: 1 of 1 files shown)


server/src/session.ts
```diff
@@ -876,33 +861,50 @@ export class Session {
    * Disable the language service if the specified `project` is not Angular or
    * Ivy mode is enabled.
    */
-  private checkProject(project: ts.server.Project) {
+  private disableLanguageServiceIfNecessary(project: ts.server.Project): void {
```
atscott, line 864: hmm, I'm not sure I like the new name for this. The function does quite a bit more than disabling if necessary (i.e. running `ngcc`). Though I do agree that `checkProject` isn't the best either. At the very least, I think this could use a javadoc update at the top.

server/src/session.ts
```diff
@@ -876,33 +861,50 @@ export class Session {
...
       // Keep language service disabled until ngcc is completed.
       project.disableLanguageService();
+      const configFilePath = project.getConfigFilePath();
       // Do not wait on this promise otherwise we'll be blocking other requests
-      this.runNgcc(project.getConfigFilePath()).catch((error: Error) => {
-        this.error(error.toString());
-      });
+      this.runNgcc(configFilePath)
+          .then(() => {
+            // Re-enable language service even if ngcc fails, because users could fix
```
atscott, line 895: Does the `.then` execute if ngcc fails?

atscott (approved):

server/src/session.ts
kyliau: The `runNgcc` function is wrapped in `try-catch-finally` so the promise it returns will never reject.

server/src/session.ts
kyliau: @atscott ok, I figured it's better to just refactor the code to remove this function entirely. PTAL.

server/src/session.ts
```diff
@@ -873,42 +842,84 @@ export class Session {
...
   /**
-   * Disable the language service if the specified `project` is not Angular or
-   * Ivy mode is enabled.
+   * Disable the language service of the specified `project` if its dependencies
+   * do not contain `@angular/core`.
+   *
+   * @returns main declaration file in `@angular/core`.
    */
-  private checkProject(project: ts.server.Project) {
+  private disableLanguageServiceIfNotAngular(project: ts.server.Project): string|undefined {
```
atscott, line 850: How about `findAngularCoreOrDisableLanguageService`?

atscott (approved):

server/src/session.ts
kyliau: sgtm, will rename and update docs.

PR #1163, opened 2021-02-27
fix: show Go to component in HTML files only

kyliau:
Go to component should not be shown in TypeScript files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,12 @@
       }
     ],
     "menus": {
+      "commandPalette": [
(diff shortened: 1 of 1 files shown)


atscott (approved):
Hmmm, I thought I did this! Thanks for the fix!

kyliau:
@atscott You set the `when` clause for the editor context (right click menu), but not the command palette (the menu that appears when you type Cmd + Shift + P)

Issue #1090, opened 2021-02-01
Ivy compiler throws error due to missing typecheck files


kyliau:
I believe this is the same issue I've been experiencing. The performance of the NGLS has been slower and slower over the past few months. It's gotten quite burdensome to use:

```
[Info  - 2:32:22 PM] Angular language server process ID: 5782
[Info  - 2:32:22 PM] Using typescript/lib/tsserverlibrary v4.1.3 from /Users/lincoln/projects/maritlabs/topdecked-unified/node_modules/typescript/lib/tsserverlibrary.js
[Info  - 2:32:22 PM] Using @angular/language-service/bundles/ivy v11.1.1 from /Users/lincoln/.vscode/extensions/angular.ng-template-11.1.1/node_modules/@angular/language-service/bundles/ivy.js
[Info  - 2:32:22 PM] Log file: /Users/lincoln/Library/Application Support/Code/logs/20201219T115842/exthost5/Angular.ng-template/nglangsvc.log
[Info  - 2:32:25 PM] Disabling language service for /Users/lincoln/projects/maritlabs/topdecked-api/tsconfig.json because it is not an Angular project ('@angular/core/core.d.ts' could not be found). If you believe you are seeing this message in error, please reinstall the packages in your package.json.
[Info  - 2:32:55 PM] Disabling language service for /Users/lincoln/projects/maritlabs/topdecked-common-util/tsconfig.json because it is not an Angular project ('@angular/core/core.d.ts' could not be found). If you believe you are seeing this message in error, please reinstall the packages in your package.json.
[Error - 2:33:23 PM] No config file for /Users/lincoln/projects/maritlabs/topdecked-common-util/src/lang/objects.test.ts

*** ... about 400 lines of these "No config file" messages. ***


[Error - 10:34:47 PM] No config file for /Users/lincoln/projects/maritlabs/topdecked-api/src/lib/routes/cards/recommendation/recommendation.routes.ts
[33mWarning:[0m Entry point '@rx-angular/template' contains deep imports into '/Users/lincoln/projects/maritlabs/topdecked-unified/node_modules/rxjs/internal/types'. This is probably not a problem, but may cause the compilation of entry points to be out of order.
[Info  - 10:35:11 PM] Enabling Ivy language service for /Users/lincoln/projects/maritlabs/topdecked-unified/tsconfig.json.
[...]

Issue #1030, opened 2020-12-15
Ivy server crashes when ngcc is run outside of language service


kyliau:
**Describe the bug**

When the extension is running, the server crashes when ngcc is run from the terminal.

**To Reproduce**

Open an Angular project, make sure Ivy language service is enabled, wait for server to be ready.
Go to terminal, run ngcc.
Server crashes.

**Expected behavior**

Running ngcc outside of extension should not crash the server.

**Logs**

```
/node_modules/@angular/language-service/bundles/ivy.js:145
            throw new Error(`Program does not contain "${fileName}" - available files are ${program.getSourceFiles().map(sf => sf.fileName).join(', ')}`);
            ^

Error: Program does not contain "/integration/project/app/app.component.ngtypecheck.ts" - available files are 
  /node_modules/typescript/lib/lib.es6.d.ts, 
  /node_modules/typescript/lib/lib.es5.d.ts, 
  /node_modules/typescript/lib/lib.es2015.d.ts, 
  /node_modules/typescript/lib/lib.dom.d.ts, 
  /node_modules/typescript/lib/lib.dom.iterable.d.ts,
  /node_modules/typescript/lib/lib.webworker.importscripts.d.ts, 
  /node_modules/typescript/lib/lib.scripthost.d.ts, 
  /node_modules/typescript/lib/lib.es2015.core.d.ts, 
  /node_modules/typescript/lib/lib.es2015.collection.d.ts,
  /node_modules/typescript/lib/lib.es2015.generator.d.ts,
  /node_modules/typescript/lib/lib.es2015.iterable.d.ts,
  /node_modules/typescript/lib/lib.es2015.promise.d.ts,
  /node_modules/typescript/lib/lib.es2015.proxy.d.ts,
  /node_modules/typescript/lib/lib.es2015.reflect.d.ts,
  /node_modules/typescript/lib/lib.es2015.symbol.d.ts,
  /node_modules/typescript/lib/lib.es2015.symbol.wellknown.d.ts,
  /integration/project/node_modules/rxjs/internal/Subscription.d.ts,
  /integration/project/node_modules/rxjs/internal/types.d.ts,
  /integration/project/node_modules/rxjs/internal/Subscriber.d.ts,
  /integration/project/node_modules/rxjs/internal/Operator.d.ts,
  /integration/project/node_modules/rxjs/internal/observable/iif.d.ts,
  /integration/project/node_modules/rxjs/internal/observable/throwError.d.ts,
[...]

PR #1164, opened 2021-02-27
fix: retain typecheck files after project reload

kyliau:
It turns out typecheck files are not retained even after the fix in
https://github.com/angular/angular/pull/40162.

For a complete explanation of the situation, see prior issue:
https://github.com/angular/vscode-ng-language-service/issues/1030

This is because even after adding the typecheck files to the list of
files to retain, tsserver will still remove them from a project if they
do not exist on disk:
https://github.com/microsoft/TypeScript/blob/3c32f6e154ead6749b76ec9c19cbfdd2acad97d6/src/server/editorServices.ts#L2188-L2193

In order to force tsserver to retain the files, we need to fake their
existence. This is done in the `ServerHost`.

Admittedly, this is not a great fix because it relies on the knowledge
of the typecheck file suffix, which is an implementation detail in the compiler.

The alternative is to explore the concept of "dynamic files", which are
files that exist only in memory and not on disk. This would require
changes to the Ivy compiler and `@angular/language-service`.
Ideally, the compiler should ask `@angular/language-service` how to name the
typecheck files, so that the behavior is dictated by LS rather than
the compiler.

fix https://github.com/angular/vscode-ng-language-service/issues/1090

### DIFF
--- a/server/src/server_host.ts
+++ b/server/src/server_host.ts
@@ -65,6 +65,11 @@ export class ServerHost implements ts.server.ServerHost {
   }
 
   fileExists(path: string): boolean {
+    // When a project is reloaded (due to changes in node_modules for example),
+    // the typecheck files ought to be retained. However, if they do not exist
+    // on disk, tsserver will remove them from project. See
+    // https://github.com/microsoft/TypeScript/blob/3c32f6e154ead6749b76ec9c19cbfdd2acad97d6/src/server/editorServices.ts#L2188-L2193
+    // To fix this, we fake the existence of the typecheck files.
     if (path.endsWith('.ngtypecheck.ts')) {
       return true;
     }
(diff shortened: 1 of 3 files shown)


server/src/server_host.ts
```diff
@@ -65,6 +65,9 @@ export class ServerHost implements ts.server.ServerHost {
   }
 
   fileExists(path: string): boolean {
+    if (path.endsWith('.ngtypecheck.ts')) {
```
atscott, line 68: Can you add a quick comment here noting why this is necessary? Without the context of the PR, it would be difficult to know why there's a special case in this function for typecheck files.

atscott (approved):

server/src/server_host.ts
kyliau: comment added

PR #1166, opened 2021-03-01
feat: Add codelens with a link to go to the component from a template

atscott:
This commit adds a codelens which provides a link to the component(s) of
an external template.

image

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -144,6 +144,8 @@ export class Session {
     conn.onCompletionResolve(p => this.onCompletionResolve(p));
     conn.onRequest(GetComponentsWithTemplateFile, p => this.onGetComponentsWithTemplateFile(p));
     conn.onRequest(GetTcbRequest, p => this.onGetTcb(p));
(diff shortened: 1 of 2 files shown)


server/src/session.ts
```diff
@@ -188,6 +190,33 @@ export class Session {
     return results;
   }
 
+  private onCodeLens(params: lsp.CodeLensParams): lsp.CodeLens[]|undefined {
+    if (!params.textDocument.uri.endsWith('.html')) {
+      return undefined;
+    }
+    let topOfDocument: lsp.Range = {start: {line: 0, character: 0}, end: {line: 0, character: 0}};
```
kyliau, line 197: use const? Also I think calling `lsp.Range.create()` is better.

server/src/session.ts
```diff
@@ -188,6 +190,33 @@ export class Session {
...
+
+
+    let codeLens: lsp.CodeLens = {
```
kyliau, line 200: please use const

kyliau (approved):

PR #1179, opened 2021-03-08
build: update the lock closed sha to the latest sha

josephperrott:
Update to use the latest lock closed sha, additionally use the full length sha for
referencing the exact version to use.


kyliau (approved):
Thanks for the quick fix!

PR #1180, opened 2021-03-08
perf: prevent requests from being sent to the server when outside Angular context

### LINKED ISSUE #1176: When Ivy is enabled, TypeScript code is super slow
**Describe the bug**

TypeScript autocompletion and navigation is super slow to a point where normal work is not anymore possible.
`Loading...` tooltips and broken autocompletion complicates working with TypeScript so much that at some point day-to-day work is exhausting.

**To Reproduce**

1. Have a "big angular" project...
2. Create a very simple file that exports one thing and import "2 or 3 things" like in videos below
3. Try to use `cmd|ctrl` + `click` or use autocompletion / IntelliSense

**Expected behavior**

Navigation after `cmd|ctrl` + `click` should be "quick".
IntelliSense should be "quick".
There should NOT be any `Loading...` tooltip that lasts forever.

**Logs**

> Log file gives us deep insight into the behavior and performance of the extension, so we strongly
> recommend you upload it.
> 
> Go to the command palette (Cmd+Shift+P on Mac, Ctrl+Shift+P on Windows), search `Open Angular Server log`
> 

I cannot provide logs because they reveal project structure, files names and I cannot take the risk of exposing such informations, sorry 😓

**Videos**

When `ivy` **disabled**

https://user-images.githubusercontent.com/1313605/110027837-c745a880-7d32-11eb-838a-a4ab9947eca4.mp4

When `ivy` **enabled**

https://user-images.githubusercontent.com/1313605/110027910-e0e6f000-7d32-11eb-8038-6f80a9771f50.mp4

**Additional context**

- Issue Type: `Bug`
- Extension Name: `ng-template`
- Extension Version: `11.2.5`
- OS Version: `Darwin x64 19.6.0`
- VSCode version: `1.54.0`

**Related issues**

- https://github.com/angular/vscode-ng-language-service/issues/1104
- https://github.com/angular/vscode-ng-language-service/issues/1168

atscott:
…gular context

The Angular Language Service extension of vscode has several pieces to
it:

* The client, which communicates directly with vscode
* The server, which handles LSP requests over an lsp connection
* @angular/language-service which provides concrete answers to LS queries

We added an optimization the @angular/language-service which exits early
when a request is outside the Angular context. This prevents unnecessary
Angular analysis of a file when we know from the start that there are no
Angular-specific results at a location.

This commit provides an additional optimization by adding a similar preventing
short-circuit from the client side. This prevents requests from even
being sent to the server when we know there is no Angular information at
a location. This optimization is a necessary addition because the server
can be blocked from processing requests if another one is taking a while
to respond. We found that requests for diagnostics, which are necessary
when opening a new file, block other server requests from being
processed. This commit would prevent that block from happening by never
making the request to the server in the first place.

fixes #1176

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,8 +7,12 @@
       "request": "launch",
       "name": "Launch Client",
       "runtimeExecutable": "${execPath}",
-      "args": ["--extensionDevelopmentPath=${workspaceRoot}"],
-      "outFiles": ["${workspaceRoot}/dist/client/*.js"],
+      "args": [
+        "--extensionDevelopmentPath=${workspaceFolder}"
+      ],
+      "outFiles": [
+        "${workspaceFolder}/dist/client/*.js"
+      ],
       "preLaunchTask": {
         "type": "npm",
         "script": "watch"
@@ -20,7 +24,19 @@
       "name": "Attach to Server",
       "port": 6009,
       "restart": true,
-      "outFiles": ["${workspaceRoot}/dist/server/*.js"]
+      "outFiles": [
+        "${workspaceFolder}/dist/server/*.js"
+      ]
+    },
+    {
+      "type": "node",
+      "request": "attach",
+      "name": "Attach to Jasmine",
+      "port": 9229,
+      "restart": true,
+      "outFiles": [
+        "${workspaceFolder}/dist/**/*.js"
+      ],
     },
     {
       "name": "Integration test: Attach to server",
@@ -29,7 +45,9 @@
       "skipFiles": [
         "<node_internals>/**"
       ],
-      "outFiles": ["${workspaceRoot}/dist/integration/lsp/*.js"],
+      "outFiles": [
+        "${workspaceFolder}/dist/integration/lsp/*.js"
+      ],
       "type": "node"
     },
     {
@@ -38,11 +56,13 @@
       "request": "launch",
       "runtimeExecutable": "${execPath}",
       "args": [
-        "--extensionDevelopmentPath=${workspaceRoot}",
-        "--extensionTestsPath=${workspaceRoot}/dist/integration/e2e",
-        "${workspaceRoot}/integration/project"
+        "--extensionDevelopmentPath=${workspaceFolder}",
+        "--extensionTestsPath=${workspaceFolder}/dist/integration/e2e",
+        "${workspaceFolder}/integration/project"
       ],
-      "outFiles": ["${workspaceRoot}/dist/integration/e2e/*.js"],
+      "outFiles": [
+        "${workspaceFolder}/dist/integration/e2e/*.js"
+      ],
       "preLaunchTask": {
         "type": "npm",
         "script": "compile:integration"
@@ -52,7 +72,10 @@
   "compounds": [
     {
       "name": "Client + Server",
-      "configurations": ["Launch Client", "Attach to Server"]
+      "configurations": [
+        "Launch Client",
+        "Attach to Server"
+      ]
     }
   ]
 }
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -15,6 +15,7 @@ import {ProjectLoadingFinish, ProjectLoadingStart, SuggestIvyLanguageService, Su
 import {NgccProgress, NgccProgressToken, NgccProgressType} from '../common/progress';
 import {GetComponentsWithTemplateFile, GetTcbRequest} from '../common/requests';
 
+import {isInsideComponentDecorator, isInsideInlineTemplateRegion} from './embedded_support';
 import {ProgressReporter} from './progress-reporter';
 
 interface GetTcbResponse {
@@ -50,6 +51,37 @@ export class AngularLanguageClient implements vscode.Disposable {
       // Don't let our output console pop open
       revealOutputChannelOn: lsp.RevealOutputChannelOn.Never,
       outputChannel: this.outputChannel,
+      middleware: {
+        provideDefinition: async (
+            document: vscode.TextDocument, position: vscode.Position,
+            token: vscode.CancellationToken, next: lsp.ProvideDefinitionSignature) => {
+          if (isInsideComponentDecorator(document, position)) {
+            return next(document, position, token);
+          }
+        },
+        provideTypeDefinition: async (
+            document: vscode.TextDocument, position: vscode.Position,
+            token: vscode.CancellationToken, next) => {
+          if (isInsideInlineTemplateRegion(document, position)) {
+            return next(document, position, token);
+          }
+        },
+        provideHover: async (
+            document: vscode.TextDocument, position: vscode.Position,
+            token: vscode.CancellationToken, next: lsp.ProvideHoverSignature) => {
+          if (isInsideInlineTemplateRegion(document, position)) {
+            return next(document, position, token);
+          }
+        },
+        provideCompletionItem: async (
+            document: vscode.TextDocument, position: vscode.Position,
+            context: vscode.CompletionContext, token: vscode.CancellationToken,
+            next: lsp.ProvideCompletionItemsSignature) => {
+          if (isInsideInlineTemplateRegion(document, position)) {
+            return next(document, position, context, token);
+          }
+        }
+      }
     };
   }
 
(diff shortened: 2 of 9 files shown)


.vscode/launch.json
```diff
@@ -20,7 +24,19 @@
...
       "restart": true,
-      "outFiles": ["${workspaceRoot}/dist/server/*.js"]
+      "outFiles": [
+        "${workspaceRoot}/dist/server/*.js"
+      ]
+    },
+    {
+      "type": "node",
+      "request": "attach",
+      "name": "Attach to Jasmine Client",
```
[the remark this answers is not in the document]
atscott: I actually made this for debugging the server and client tests though I suppose it would also work for the LSP tests. It's simply set up to attach to the default port used by jasmine.

client/tsconfig.json
```diff
@@ -1,12 +1,23 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
+    "composite": true,
     "outDir": "../dist/client",
-    "rootDirs": [".", "../dist"]
+    "rootDir": "src",
```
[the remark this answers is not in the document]
atscott: Rollup was having problems when it wasn't there: `Error: Could not resolve entry module (dist/client/extension.js).`

client/src/client.ts
```diff
@@ -50,6 +51,36 @@ export class AngularLanguageClient implements vscode.Disposable {
       // Don't let our output console pop open
       revealOutputChannelOn: lsp.RevealOutputChannelOn.Never,
       outputChannel: this.outputChannel,
+      middleware: {
+        provideDefinition: async (document, position, token, next) => {
+          if (document.languageId === 'typescript' &&
+              !isInsideAngularContext(document.getText(), document.offsetAt(position))) {
```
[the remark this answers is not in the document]
atscott: Added the explicit types to the parameters. `position` is a `vscode.Position` while the `offsetAt` converts it to a number.

client/src/client.ts
```diff
@@ -50,6 +51,36 @@ export class AngularLanguageClient implements vscode.Disposable {
       // Don't let our output console pop open
       revealOutputChannelOn: lsp.RevealOutputChannelOn.Never,
       outputChannel: this.outputChannel,
+      middleware: {
+        provideDefinition: async (document, position, token, next) => {
```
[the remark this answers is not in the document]
atscott: I'd like to keep it this way if that's alright with you. If/when we implement virtual documents in the middleware to help resolve some of the open requests with inline templates, I'll be tracking the virtual document contents in a uri to vdoc contents `Map` that lives on the `AngularLanguageClient`. Also, autocomplete for the function name works when it's this form but not `async <function>`.

client/src/embedded_support.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import * as ts from 'typescript';
```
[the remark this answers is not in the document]
atscott: Yes, the startup time before appears to be 200-300ms and the startup time now looks like 500-600ms. So 2-3x slower.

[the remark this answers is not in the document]
atscott: I've done some experimentation based on the regex in the [syntaxes](https://github.com/angular/vscode-ng-language-service/blob/master/syntaxes/inline-template.json#L10) and it seems like something that would be feasible without introducing a dependency on typescript: https://github.com/atscott/vscode-ng-language-service/blob/c1b86ff50dd9b4e8c9b29c4e71127483ee26b0e6/client/src/embedded_support.ts

client/src/client.ts
```diff
@@ -15,6 +15,7 @@ import {ProjectLoadingFinish, ProjectLoadingStart, SuggestIvyLanguageService, Su
 import {NgccProgress, NgccProgressToken, NgccProgressType} from '../common/progress';
 import {GetComponentsWithTemplateFile, GetTcbRequest} from '../common/requests';
 
+import {isInsideAngularContext, isInsideInlineTemplateRegion} from './embedded_support';
```
kyliau, line 18: nit: I think `isInsideComponentDecorator` is clearer, but I'll let you make the call

client/src/client.ts
```diff
@@ -50,6 +51,45 @@ export class AngularLanguageClient implements vscode.Disposable {
       // Don't let our output console pop open
       revealOutputChannelOn: lsp.RevealOutputChannelOn.Never,
       outputChannel: this.outputChannel,
+      middleware: {
+        provideDefinition: async (
+            document: vscode.TextDocument, position: vscode.Position,
+            token: vscode.CancellationToken, next: lsp.ProvideDefinitionSignature) => {
+          if (document.languageId === 'typescript' &&
```
kyliau, line 58: What do you think if we simplify this to:
```ts
if (isInsideInlineTemplateRegion(document)) {
  next(document, position, token);
}
```
and do all the checks in the function?

client/src/embedded_support.ts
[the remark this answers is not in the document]
kyliau: As discussed, we'll proceed with this change, but note the regression in activation time in the change log.

client/src/embedded_support.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  let token: ts.SyntaxKind = scanner.scan();
+  let lastToken: ts.SyntaxKind|undefined;
+  let lastTokenText: string|undefined;
+  let unclosedBraces = 0;
+  let unclosedBrackets = 0;
+  let propertyAssignmentContext = false;
+  while (token !== ts.SyntaxKind.EndOfFileToken && scanner.getStartPos() < offset) {
+    if (lastToken === ts.SyntaxKind.Identifier &&
+        propertyAssignmentNames.some(name => name === lastTokenText) &&
```
kyliau, line 46:
```suggestion
        propertyAssignmentNames.includes(lastTokenText) &&
```
```suggestion
        propertyAssignmentNames.some(name => name === lastTokenText) &&
```

client/src/embedded_support.ts
```diff
@@ -0,0 +1,88 @@
...
+      continue;
+    }
+    if (unclosedBraces === 0 && unclosedBrackets === 0 && isPropertyAssignmentTerminator(token)) {
+      propertyAssignmentContext = false;
+    }
+
+    if (token === ts.SyntaxKind.OpenBracketToken) {
+      unclosedBrackets++;
+    }
+    if (token === ts.SyntaxKind.OpenBraceToken) {
```
kyliau, line 59:
```suggestion
    else if (token === ts.SyntaxKind.OpenBraceToken) {
```
```suggestion
    if (token === ts.SyntaxKind.OpenBraceToken) {
```

client/src/embedded_support.ts
```diff
@@ -0,0 +1,88 @@
...
+      unclosedBraces++;
+    }
+    if (token === ts.SyntaxKind.CloseBracketToken) {
```
kyliau, line 62:
```suggestion
    else if (token === ts.SyntaxKind.CloseBracketToken) {
```
```suggestion
    if (token === ts.SyntaxKind.CloseBracketToken) {
```

client/src/embedded_support.ts
```diff
@@ -0,0 +1,88 @@
...
+      unclosedBrackets--;
+    }
+    if (token === ts.SyntaxKind.CloseBraceToken) {
```
kyliau, line 65:
```suggestion
    else if (token === ts.SyntaxKind.CloseBraceToken) {
```
```suggestion
    if (token === ts.SyntaxKind.CloseBraceToken) {
```

client/src/embedded_support.ts
```diff
@@ -0,0 +1,88 @@
...
+      unclosedBraces--;
+    }
+
+    const isStringToken = token === ts.SyntaxKind.StringLiteral ||
+        token === ts.SyntaxKind.NoSubstitutionTemplateLiteral;
+    const isCursorInToken = scanner.getStartPos() <= offset &&
+        scanner.getStartPos() + scanner.getTokenText().length >= offset;
```
kyliau, line 72: Should the `offset` be strictly less than `scanner.getStartPos() + scanner.getTokenText().length`?

kyliau (approved):

client/src/embedded_support.ts
atscott: Added test with comment about why we include that position.

kyliau (approved):

PR #1181, opened 2021-03-08
perf: Support request cancellation

kyliau:
When multiple requests of the same type are sent in quick succession, the
LSP client will automatically issue `$/cancelRequest` notification to cancel
the earlier requests.

For example, when the mouse is moving around, multiple `hover` requests could
be fired, but only the last one is meaningful to users. In this case, we
should handle the cancelled requests and not do any work.

This will significantly speed up our server's ability to process incoming
requests / notifications.


atscott (approved):
🐎 ✅

Issue #1191, opened 2021-03-10
Allow users to hide strictTemplates prompt


ehavener:
Referencing **#1053 Prompt users to enable strictTemplates** 

When an Angular project contains many `.tsconfig` project roots that extend a 'base' `.tsconfig`, this prompt appears every time one of them is loaded. When working in an [@nrwl/nx](https://github.com/nrwl/nx) workspace, with separate projects defined for every module, these notifications appear constantly. 

This is because the language service is enabled per each 'sub' project. Diagnostics then emit per each language service. 

When we get a diagnostic relating to `compilerOptions` on a parent root, and this parent's config is unchanged since a sibling's last diagnostic, we can disregard that diagnostic.

There are other cases where we could make the client smarter about sibling projects that might be beneficial. But this sounds like a rather tall order. For example, there doesn't seem to be a public way to get the path of the 'base' `.tsconfig` short of compiling all paths manually. `rootDir`, `baseUrl`, etc. are unreliable and have different uses.

Anyways, getting back to the point, can we **allow the user to disable this prompt**?

![Screen Shot 2021-03-10 at 2 48 51 PM](https://user-images.githubusercontent.com/18430808/110708630-d05dcc00-81af-11eb-830d-697bbee41495.png)

kyliau:
I wonder if vscode has built in capability to permanently hide a notification instead of us handling it in the extension.
I've seen screenshots like this (from https://github.com/microsoft/vscode/issues/94923):
![image](https://user-images.githubusercontent.com/2941178/110718216-651bf600-81bf-11eb-9bb1-da6bb0eedf09.png)
Does clicking on the cog wheel icon present the "Do not show this again" option (provided by vscode, not this extension)?
I think it might be a Windows-only thing, because on Mac the cog wheel doesn't show up at all:
<img width="1792" alt="Screen Shot 2021-03-10 at 4 43 28 PM" src="https://user-images.githubusercontent.com/2941178/110718418-cba11400-81bf-11eb-9a1a-4e0ff3c3e958.png">

ehavener:
It does:
![Screen Shot 2021-03-10 at 4 48 03 PM](https://user-images.githubusercontent.com/18430808/110718741-6e599280-81c0-11eb-874c-449b15eccc4d.png)
however I've only seen the "Do not show this again" option on alerts direct from vscode. I don't think there's an API for it.

If we want to avoid a setting, we could use one of these methods to store meta to hide it
[...]

PR #1193, opened 2021-03-10
feat: Allow users to hide strictTemplates prompt

ehavener:
https://github.com/angular/vscode-ng-language-service/issues/1191

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -173,19 +173,29 @@ function registerNotificationHandlers(
 
   const disposable2 =
       client.onNotification(SuggestStrictMode, async (params: SuggestStrictModeParams) => {
+        const config = vscode.workspace.getConfiguration();
+        if (config.get('angular.enable-strict-mode-prompt') === false) {
+          return;
+        }
+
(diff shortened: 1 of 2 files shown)


ehavener:
@googlebot I signed it!

client/src/client.ts
```diff
@@ -173,19 +173,29 @@ function registerNotificationHandlers(
...
                 '[angularCompilerOptions](https://angular.io/guide/angular-compiler-options).',
             openTsConfig,
+            doNotPromptAgain,
         );
         if (selection === openTsConfig) {
           const document = await vscode.workspace.openTextDocument(params.configFilePath);
           vscode.window.showTextDocument(document);
+        } else if (selection === doNotPromptAgain) {
+          config.update(
+              'angular.enable-strict-mode-prompt', false, vscode.ConfigurationTarget.Global);
```
kyliau, line 198: I think this should be per workspace rather than global, i.e. it should still prompt users to turn on strict mode if they switch to another project.

kyliau (approved):

kyliau:
Thanks for fixing this!

PR #1239, opened 2021-03-31
Create node.js.yml for E2E test on GitHub Actions

kyliau:
Run E2E test on Linux and MacOS.
(Test currently fails on Windows, further debugging required).

Ref https://code.visualstudio.com/api/working-with-extensions/continuous-integration

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,0 +1,30 @@
+# This workflow will do a clean install of node dependencies, build the source code and run the E2E tests.
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: E2E Test
+
+on:
+  push:
+    branches: [ master ]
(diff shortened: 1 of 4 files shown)


atscott (approved):

kyliau:
For the record, test on Windows failed with error code 3765269347
Here's the output from the run: https://github.com/angular/vscode-ng-language-service/runs/2239507564
Similar issue https://travis-ci.community/t/vscode-extension-builds-failing-exit-code-3221225781/7187 apparently caused by missing DLLs.

kyliau:
@josephperrott We did the migration already.
In #1222 the type imports from `vscode` package were replaced with `@types/vscode`, and
in #1224 the test runner from `vscode` was replaced with that from `vscode-test`.

josephperrott:
@kyliau I was seeing its use in the `definition_spec.ts` file, looking closer now, I see its a different usage.

PR #1246, opened 2021-03-31
build: add configuration to launch prod client

kyliau:
Add a command to launch a dev instance that uses the production .vsix.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,7 +5,7 @@
     {
       "type": "extensionHost",
       "request": "launch",
-      "name": "Launch Client",
+      "name": "Launch Dev Client",
       "runtimeExecutable": "${execPath}",
       "args": [
         "--disable-extensions",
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -19,6 +19,18 @@
         "script": "watch"
       }
     },
+    {
+      "type": "extensionHost",
+      "request": "launch",
+      "name": "Launch Prod Client",
+      "runtimeExecutable": "${execPath}",
+      "args": [
+        "--disable-extensions",
```
atscott, line 28: May I suggest maybe removing the `--disable-extensions` from either the dev client or the prod client? Or maybe adding yet another config to include extensions for debugging g3?

atscott (approved):

.vscode/launch.json
kyliau: Yeah, I think it makes sense to not disable extensions when testing the prod build, since we'll most likely use this config to test g3.

PR #1259, opened 2021-04-05
perf: Avoid making Angular-related decisions for files not in an Angu…

### LINKED ISSUE #1237: Implement way for language server to communicate to client that Angular is not in workspace
We currently have middleware in the client side that tokenizes typescript files to determine if the position is within an inline template before forwarding the request to the server. This is done for performance reasons and allows us to prevent the server from being a bottleneck for typescript files in these cases. However, this tokenization is done for all typescript files, even in non-Angular projects. We should implement a way to communicate from the server back to the client that Angular was not detected in the workspace and that we should skip the tokenization/short-circuit logic.

atscott:
…lar project

This commit updates our client-side short-circuit logic to provide an
additiona short circuit that avoids tokenizing typescript files when
they are not within a project that uses Angular.

fixes #1237

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -13,7 +13,7 @@ import * as lsp from 'vscode-languageclient/node';
 
 import {ProjectLoadingFinish, ProjectLoadingStart, SuggestIvyLanguageService, SuggestIvyLanguageServiceParams, SuggestStrictMode, SuggestStrictModeParams} from '../common/notifications';
 import {NgccProgress, NgccProgressToken, NgccProgressType} from '../common/progress';
-import {GetComponentsWithTemplateFile, GetTcbRequest} from '../common/requests';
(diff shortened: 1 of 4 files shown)


client/src/client.ts
```diff
@@ -117,6 +123,23 @@ export class AngularLanguageClient implements vscode.Disposable {
...
+  private async isInAngularProject(doc: vscode.TextDocument): Promise<boolean> {
+    if (this.client === null) {
+      return false;
+    }
+    const uri = doc.uri.toString();
+    if (this.fileToIsInAngularProjectMap.has(uri)) {
+      return this.fileToIsInAngularProjectMap.get(uri)!;
+    }
+
+    const response = await this.client.sendRequest(IsInAngularProject, {
```
kyliau, line 135: Please add try-catch to handle errors, otherwise it'll crash the client.

kyliau (approved):

PR #1260, opened 2021-04-05
fix: do not resolve CodeLens command until LS is enabled

### LINKED ISSUE #1253: CodeLens show !!MISSING: command!! on startup
Screen Shot 2021-03-31 at 3 45 46 PM

I don't remember how to reproduce this exactly, but I saw it on startup. It fixes itself after a while:

Screen Shot 2021-03-31 at 3 45 53 PM

atscott:
CodeLens will show "!!MISSING: command!!" on startup while ngcc is
running (because the language service gets disabled). Fortunately, the
CodeLensResolve gets called periodically so we can simply wait until the
language service is re-enabled before returning the resolved command.

Fixes #1253

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -228,6 +228,16 @@ export class Session {
   }
 
   private onCodeLensResolve(params: lsp.CodeLens): lsp.CodeLens {
+    const lsInfo = this.getLSAndScriptInfo(params.data);
+    if (lsInfo === undefined) {
+      return params;
+    }
+    const project = this.getDefaultProjectForScriptInfo(lsInfo.scriptInfo);
+    // If the language service is disabled, the angular command will not be available.
+    if (!project?.languageServiceEnabled) {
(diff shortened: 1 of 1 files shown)


server/src/session.ts
```diff
@@ -228,6 +228,16 @@ export class Session {
   }
 
   private onCodeLensResolve(params: lsp.CodeLens): lsp.CodeLens {
+    const lsInfo = this.getLSAndScriptInfo(params.data);
+    if (lsInfo === undefined) {
+      return params;
+    }
+    const project = this.getDefaultProjectForScriptInfo(lsInfo.scriptInfo);
+    // If the language service is disabled, the angular command will not be available.
+    if (!project || !project.languageServiceEnabled) {
```
kyliau, line 237:
```suggestion
    if (!project?.languageServiceEnabled) {
```

kyliau (approved):

PR #1264, opened 2021-04-06
fix: mark project as dirty after a TS file is opened

kyliau:
When a TS file is opened, its script version is changed because
the internal representation switches from TextStorage to ScriptVersionCache.
This version change will cause the source file to be updated and
invalidate the source file cached by NgCompiler.

With this change, we mark the project as dirty after opening a TS file
to force a new program to be created, thereby triggering NgCompiler
to reanalyze the program and pick up the changed source file.


atscott (approved):

atscott (approved):

kyliau:
Blocked on Alex's changes to compiler https://github.com/angular/angular/pull/41475

kyliau:
Closing in favor of https://github.com/angular/angular/pull/41500

PR #1277, opened 2021-04-12
feat: add support for signature help

alxhub:
This commit adds support for signature help to the LS extension, by
translating from the TS `getSignatureHelpItems` API into LSP.

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -95,6 +95,15 @@ export class AngularLanguageClient implements vscode.Disposable {
               await Promise.all([angularResultsPromise, htmlProviderResultsPromise]);
           return angularResults ?? htmlProviderResults?.[0];
         },
+        provideSignatureHelp: async (
+            document: vscode.TextDocument, position: vscode.Position,
+            context: vscode.SignatureHelpContext, token: vscode.CancellationToken,
+            next: lsp.ProvideSignatureHelpSignature) => {
+          if (await this.isInAngularProject(document) &&
+              isInsideInlineTemplateRegion(document, position)) {
+            return next(document, position, context, token);
+          }
+        },
         provideCompletionItem: async (
             document: vscode.TextDocument, position: vscode.Position,
             context: vscode.CompletionContext, token: vscode.CancellationToken,
--- a/integration/lsp/ivy_spec.ts
+++ b/integration/lsp/ivy_spec.ts
@@ -155,6 +155,61 @@ describe('Angular Ivy language server', () => {
     });
   });
 
+  describe('signature help', () => {
+    it('should show signature help for an empty call', async () => {
+      client.sendNotification(lsp.DidOpenTextDocumentNotification.type, {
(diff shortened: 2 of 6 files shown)


zarend (approved):
Looks good, but how come there are no tests? :)

alxhub:
I can't seem get the tests to pass on my machine. They fail even on `master`, with a fresh `node_modules`:

```
$ yarn test
yarn run v1.21.1
warning ng-template@12.0.0-next.0: The engine "vscode" appears to be invalid.
$ yarn compile:test && jasmine --config=jasmine.json
warning ng-template@12.0.0-next.0: The engine "vscode" appears to be invalid.
$ tsc -b server/src/tests && tsc -b client/src/tests
Randomized with seed 92707
Started
.........................F..........

Failures:
1) resolveAndRunNgcc runs ngcc from node_modules where ngcc is resolved to
  Message:
    Expected [object Promise] to be resolved.
  Stack:
    Error
        at <Jasmine>
        at UserContext.<anonymous> (/Users/arick/dev/vscode-ng-language-service/dist/server/tests/ngcc_spec.js:25:36)
        at <Jasmine>
        at processImmediate (internal/timers.js:439:21)
  Message:
    Expected 0 to be 1.
  Stack:
    Error: Expected 0 to be 1.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/arick/dev/vscode-ng-language-service/dist/server/tests/ngcc_spec.js:26:35)
  Message:
    TypeError: Cannot destructure property `cwd` of 'undefined' or 'null'.
  Stack:
    TypeError: Cannot destructure property `cwd` of 'undefined' or 'null'.
        at UserContext.<anonymous> (/Users/arick/dev/vscode-ng-language-service/dist/server/tests/ngcc_spec.js:29:15)

36 specs, 1 failure
Finished in 0.043 seconds
```

So I want to add tests in a followup, after I talk to @kyliau and figure out what's up here.

atscott (changes requested):
I think we should also add the client side short-circuit for this operation (ex: https://github.com/angular/vscode-ng-language-service/blob/master/client/src/client.ts#L67-L70)

alxhub:
> I think we should also add the client side short-circuit for this operation (ex: https://github.com/angular/vscode-ng-language-service/blob/master/client/src/client.ts#L67-L70)

Good call, done.

alxhub:
> So I want to add tests in a followup

I actually think this is necessary, since the framework LS PR hasn't landed yet. The feature can't actually work until that functionality is landed, released in an Angular version, and the extension is updated to depend on that version.

atscott (approved):

alxhub:
Per a conversation with @atscott, the practice for the extension repo is to open a PR with tests even if they're breaking, and only merge it once the feature PR lands upstream.

Tests are now added :)

kyliau:
blocked until 12.0.0-next.9 release

PR #1279, opened 2021-04-12
fix: make Ivy LS the default

kyliau:
Ivy LS will be the default in Angular v12.

To revert to legacy View Engine LS, enable the `view-engine` option:

Screen Shot 2021-04-12 at 2 12 26 PM


atscott (approved):
LGTM (pending rebase)

PR #1367, opened 2021-05-19
test: quality of life updates for e2e tests

atscott:
- update vscode version
- use __dirname for resolving test file locations to be more fault tolerant of potential errors in directory resolution
- wait for definitions to be available during activation rather than waiting a set time


integration/e2e/jasmine.ts
```diff
@@ -4,12 +4,14 @@ export async function run(): Promise<void> {
   const jasmine = new Jasmine({});
```
kyliau, line 4: Should specify `projectBaseDir` here?

kyliau (approved):

integration/e2e/jasmine.ts
atscott: 👍

PR #1448, opened 2021-07-14
test: update e2e test wait condition to wait for definitions

atscott:
Rather than waiting a set 20 seconds for ngcc to finish, this updates the tests to
instead wait for definitions to be available in the template. This change was intended
to be in #1367 but accidentally was reverted before merging

### DIFF
--- a/integration/e2e/helper.ts
+++ b/integration/e2e/helper.ts
@@ -13,5 +13,21 @@ function sleep(ms: number) {
 
 export async function activate(uri: vscode.Uri) {
   await vscode.window.showTextDocument(uri);
-  await sleep(20 * 1000);  // Wait for server activation, including ngcc run
+  await waitForDefinitionsToBeAvailable(20);
+}
+
+async function waitForDefinitionsToBeAvailable(maxSeconds: number) {
+  let tries = 0
+  while (tries < maxSeconds) {
+    const position = new vscode.Position(4, 25);
+    // For a complete list of standard commands, see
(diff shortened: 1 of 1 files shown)


integration/e2e/helper.ts
```diff
@@ -13,5 +13,21 @@ function sleep(ms: number) {
...
+async function waitForDefinitionsToBeAvailable(maxSeconds: number) {
+  let tries = 0
+  while (tries < maxSeconds) {
+    const position = new vscode.Position(4, 25);
+    // For a complete list of standard commands, see
+    // https://code.visualstudio.com/api/references/commands
+    const definitions = await vscode.commands.executeCommand<vscode.LocationLink[]>(
+        DEFINITION_COMMAND, APP_COMPONENT_URI, position);
+    if (definitions && definitions.length > 0) {
+      return;
```
zarend, line 28: It might be safer to sleep for a short time here (about 20-100ms). That's in case there's sometime else we need to wait for that we don't know about. If we've already waited at least 1000ms, then we probably don't need to add another wait. ¯\_(ツ)_/¯

zarend (approved):

integration/e2e/helper.ts
atscott: This is waiting for `ngcc` to complete so it really could take a long time. There's really no knowing. That's why the timeout was 20 seconds before. This is also 20 seconds, but with a possible early exit.

PR #1463, opened 2021-07-28
test: update default timeout inteval to 30 seconds

atscott:
This change is an effort to reduce the flakiness of the e2e tests on macos.
If those e2e tests are still still flaky after this change, we should consider them for macos.

example failure due to timeout: https://github.com/angular/vscode-ng-language-service/runs/3186857720


zarend (approved):
🤞

PR #1493, opened 2021-08-26
refactor(server): Prefer `null` over `undefined` and explicit `void`

atscott:
Update return types of functions to:

* Explicitly have `void` as the return type
* Prefer `null` over `undefined`

The LSP spec always uses `null` as the return types:
https://microsoft.github.io/language-server-protocol/specifications/specification-current/

In fact, if the server is configured to return `undefined`, it will
still be `null` after coming through the LSP connection. To avoid any
confusion, we should just always use `null`.

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -332,7 +332,7 @@ export class Session {
     return params;
   }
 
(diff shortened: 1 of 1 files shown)


clydin (approved):
LGTM.
nice cleanup.

PR #1509, opened 2021-10-01
feat: provide snippets for attribute

ivanwonder:


atscott (commented):
BTW, I just tried this out and the completion experience is _really_ nice. Thanks for all your hard work!

package.json
```diff
@@ -118,6 +118,11 @@
           "type": "boolean",
           "default": true,
           "description": "Enable/disable showing completions on potentially undefined values that insert an optional chain call. Requires TS 3.7+ and strict null checks to be enabled."
+        },
+        "angular.suggest.includeCompletionsWithSnippetText": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable/disable snippet completions from angular language Server. Requires using TypeScript 4.3+ in the workspace."
```
atscott, line 125: Is the part about requiring TS 4.3+ accurate? Wouldn't the feature simply require the client to have `snippetSupport` in the capabilities? Or have I forgotten something on the `@angular/language-service` side of things that requires TS 4.3?

This feature _does_ require that the "legacy View Engine" option is not enabled though (same with `includeAutomaticOptionalChainCompletions`, but I missed that in the last review)

atscott (commented):
LGTM other than the one question about 
```
includeCompletionsWithInsertText: this.includeAutomaticOptionalChainCompletions || includeCompletionsWithSnippetText,
```

PR #1567, opened 2021-11-16
ci: setup scheduled CI job to test repository against the latest builds

josephperrott:
Use a scheduled job to check the repository against the latest version of @angular/language-service from
https://github.com/angular/language-services-builds


atscott (approved):

atscott:
Looks like you need to also update the workflow to pass in `package.json`: https://github.com/angular/vscode-ng-language-service/blob/master/.github/workflows/node.js.yml#L27

PR #1571, opened 2021-11-19
test: fix syntax tests so they work when testing against builds repo

atscott:
When running tests against the builds repo, the package.json and yarn.lock files get updated.
The syntax check only really cares about modifications to files in the syntaxes folder.


josephperrott (approved):
LGTM

PR #1589, opened 2022-01-12
feat(extension): Support renaming from TypeScript files

ivanwonder:
This PR will add a new package that is applied to the built-in TS
extension by the config [`typescriptServerPlugins`][1] and is used
to disable rename provider of the built-in TS extension so VSCode
asks the Angular Language Service for the answer instead.

Detail about this package is [here][2].

[1]: https://code.visualstudio.com/api/references/contribution-points#contributes.typescriptServerPlugins
[2]: https://github.com/angular/angular/blob/master/packages/language-service/README.md#override-rename-ts-plugin

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -65,16 +65,7 @@ export class AngularLanguageClient implements vscode.Disposable {
         prepareRename: async (
             document: vscode.TextDocument, position: vscode.Position,
             token: vscode.CancellationToken, next: lsp.PrepareRenameSignature) => {
-          // We are able to provide renames for many types of string literals: template strings,
-          // pipe names, and hopefully in the future selectors and input/output aliases. Because
-          // TypeScript isn't able to provide renames for these, we can more or less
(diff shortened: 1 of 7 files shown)


ivanwonder:
The plugin code move into this [PR](https://github.com/angular/angular/pull/44696)

atscott:
Adding the blocked label. This has to wait for the 13.2 release here since we don't yet have a branch for the next version that https://github.com/angular/angular/pull/44696 was included in.

atscott (changes requested):
@ivanwonder LGTM aside from the commit message. Like the other commit, could you write it as the feature you're providing rather than the implementation details about how that was achieved?

That is: 
```
feat(extension): Support renaming from TypeScript files

<<details about how you did this. Disable rename provider of built in 
TS extension so VSCode asks the Angular Language Service for the answer instead. >>
```

atscott (approved):
🥳

PR #1604, opened 2022-01-27
fix(server): return the right range for the original source file of DTS

ivanwonder:
When the user tries to go to the definition that is in the DTS file,
the Angular LS will map it to the original source file. Because of
not updating the `scriptInfo`, the Angular LS returns the wrong range
of the definition.


ivanwonder:
I can't find a way to add a test for it. there is no way to convert `lsp.Range` which is made up of `start` and `end` positions to TypeScript's absolute offsets out of the server. and getting the content of a file by the `fs` is not a good way.

atscott (approved):

Issue #1607, opened 2022-02-04
Progress notifications lack `kind` field


j-hui:
# 🐞 bug report

### Is this a regression?

Not to my knowledge.

### Description

I'm the author of a [Neovim plugin](https://github.com/j-hui/fidget.nvim) that reports the ongoing progress of LSP servers via the [`$/progress` endpoint](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#workDoneProgressEnd).

When using my plugin with this language server, [some users noticed](https://github.com/j-hui/fidget.nvim/pull/26#issue-1118502502) that angularls was not sending any `end` notifications. This does not appear to be compliant with the LSP spec.

I took a peek at the source code and found progress being [sent directly via the `sendProgress` method](https://github.com/angular/vscode-ng-language-service/blob/9a5ecd92cf617259e474fc23461cb586cc6ead76/server/src/session.ts#L1197) without any `kind` field. This seems to differ from the library-provided [`done` helper](https://github.com/microsoft/vscode-languageserver-node/blob/main/server/src/common/progress.ts#L72) helper, which takes cares of specifying the `end` kind.

## Bug Type

- [ ] Angular Language Service VSCode extension
- [x] Angular Language Service server

## Reproduction

I'm not able to reproduce this behavior myself since I don't really know my way around web projects, but if there is a small demo project you could point me to, I could probably easily reproduce it just by opening that.

**Expected behavior**

When progress is complete, the server sends an `end` notification.

### **Logs**

<pre><code>
n/a
</code></pre>

### **Screenshots**

See https://github.com/j-hui/fidget.nvim/pull/26

## 🌍  Your Environment

```
$ nvim --version
NVIM v0.6.1
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by nixbld

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "
/nix/store/10k3bc68v0602pvaz6m5k420q13x83g6-neovim-unwrapped-0.6.1/share/nvim
"

Run :checkhealth for more info
```

atscott:
@j-hui Thanks for the thorough explanation. If you wanted to try reproducing, the test project used here is in `integration/project`. All you would need to do is install the `node_modules` with `yarn`.

Would you like to send a PR to fix this? Seems like you've already done all the hard work in determining what the correct fix is and all that's needed.

[...]

PR #1614, opened 2022-02-27
refactor: use built-in `WorkDoneProgressReport` to report NGCC progress

ivanwonder:
Now the Angular extension implements the `WorkDoneProgressReport`
by itself on the client-side to support the multi-report progress
with different `tsconfig.json`, to show an error message, to
dispose of the report progress, but it's already a built-in feature
that can be used on the service side.

On the service side, the built-in `WorkDoneProgressReport` will
attach a unique token to the different report progress and clean
the progress when the client is closing.

On the service side, If an error report when running NGCC, the LSP
also supports showing an error message in VSCode, but the ability to
open the output channel is not supported(now the LSP only supports
to open the document URI). This PR will only need to provide it.

So the custom `WorkDoneProgressReport` can be removed on the client side.


server/src/session.ts
```diff
@@ -1164,41 +1167,47 @@ export class Session {
...
     try {
       this.projectNgccQueue.push({project, done: false});
       await resolveAndRunNgcc(configFilePath, {
         report: (msg: string) => {
-          this.connection.sendProgress(NgccProgressType, NgccProgressToken, {
-            done: false,
-            configFilePath,
-            message: msg,
-          });
+          progressReporter.report(50, msg);
```
atscott, line 1178: I'm not a huge fan of the percentage. The first message is definitely not an indicator of 50% and there's really no way to tell an accurate percent. Can we use `undefined` in the begin and omit the number parameter here?

This isn't a huge deal if you don't think that makes sense.

server/src/session.ts
```diff
@@ -1164,41 +1167,47 @@ export class Session {
...
+          }>(`Angular extension might not work correctly because ngcc operation failed. ` +
+                 `Try to invoke ngcc manually by running 'npm/yarn run ngcc'. ` +
+                 `Please see the extension output for more information.`,
+             {title: 'Show output', action: NgccErrorMessageAction.showOutput})
+          .then((selection) => {
+            switch (selection?.action) {
+              case NgccErrorMessageAction.showOutput:
+                this.connection.sendNotification(OpenOutputChannel, {});
+                break;
+            }
```
atscott, line 1200: nit: given that there is only 1 option here, I think this would be best written as an `if`
```suggestion
            if (selection?.action === NgccErrorMessageAction.showOutput) {
                this.connection.sendNotification(OpenOutputChannel, {});
            }
```

This can always change later if we really do have a bunch of options, but I don't think that's likely to happen any time soon.

atscott:
This is actually a really nice cleanup. Could you provide some more details in the commit message about how this accomplishes the same thing as before, but with much less code? That is, drop in some of the context from https://github.com/angular/vscode-ng-language-service/issues/1607#issuecomment-1053551651 and other comments in that thread.

ivanwonder:
After the tip `Initializing Angular language features` disappeared, there is a delay to show the tip `Running ngcc for ./tsconfig.json`, and I think it's acceptable.

PR #1625, opened 2022-04-08
Fix: Ensure projects stay open and are correctly prioritized for solution style configs

atscott:
note: blocked by upstream change to `@angular/language-service` to add the `getTemplateReferences` function

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -56,7 +56,6 @@ export class Session {
   private readonly logger: ts.server.Logger;
   private readonly ivy: boolean;
   private readonly disableAutomaticNgcc: boolean;
-  private readonly configuredProjToExternalProj = new Map<string, string>();
   private readonly logToConsole: boolean;
   private readonly openFiles = new MruTracker();
   private readonly includeAutomaticOptionalChainCompletions: boolean;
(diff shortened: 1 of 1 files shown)


ivanwonder:
I have a question, why not add the template to the root of the project when the Angular compiler read the resource or adds the TCB to the root.

atscott:
@ivanwonder From my understanding, the TCB files are tracked through `getExternalFiles`. As far as I can tell, this does not result in them being added as root files to the project. Do you want to experiment with some things on your end? The project I've been using to reproduce the problem is https://github.com/itspers/vscode-angular-nx. The concrete steps are:

Reproduction steps for one issue:

1. have only this file open
2. reload window
3. wait for angular information to be available (hover over vscode-angular-nx-lib-component and wait for info)
4. while hovering over lib component hold command _as if_ you were going to go to definition. This eagerly instantiates the lib project (right now, this will also the app projects to be closed because you're opening the lib project and the html file isn't enough to make the project stay open)

ivanwonder:
When the Angular compiler read the template resource, it will create `ScriptInfo` and attach it to the current project. If the `ScriptInfo` is been added to the root of the project here, there is no need to do it every time when the user opens a template, and no need to check if the template belongs to the project.

https://github.com/angular/angular/blob/d11d1c0b70bda8a275a16ebd9a192beafcbc2a14/packages/language-service/src/adapters.ts#L94-L112

the method `getScriptSnapshot` will create `ScriptInfo` and attach it to the `Project`.

PR #1659, opened 2022-05-12
test: Update integration project to target v14

atscott:
This commit updates the integration project used in tests to target the v14 rc

### DIFF
--- a/integration/lsp/ivy_spec.ts
+++ b/integration/lsp/ivy_spec.ts
@@ -143,7 +143,7 @@ describe('Angular Ivy language server', () => {
     // 2 results - the NgIf class and the ngIf input
     expect(Array.isArray(response)).toBe(true);
     const {targetUri} = response[0];
-    expect(targetUri.endsWith('angular/common/common.d.ts')).toBeTrue();
+    expect(targetUri.endsWith('angular/common/index.d.ts')).toBeTrue();
     // Open the `.d.ts` file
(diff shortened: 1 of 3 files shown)


ivanwonder:
I think the diagnostics are triggered by ngcc. ngcc generates a lockfile named `__ngcc_lock_file__` in the node_modules directory, this triggers the directory watcher, and [ProjectsUpdatedInBackgroundEvent](https://github.com/angular/vscode-ng-language-service/blob/8c809a2982ccad19997caef7ffeaa899e9d8ddb5/server/src/session.ts#L491) event is send by the method [delayUpdateProjectGraphAndEnsureProjectStructureForOpenFiles](https://github.com/microsoft/TypeScript/blob/56a4a93718ce016f038bcbce425bd3f47c3bec78/src/server/editorServices.ts#L1356).

But this [PR](https://github.com/angular/angular/pull/44228) broke this. because the `.ngcc_lock_file` is [ignored by tsserver](https://github.com/microsoft/TypeScript/blob/ab2523bbe0352d4486f67b73473d2143ad64d03d/src/compiler/sys.ts#L500) and the tsserver will [skip updating the project](https://github.com/microsoft/TypeScript/blob/56a4a93718ce016f038bcbce425bd3f47c3bec78/src/server/editorServices.ts#L1316).

I think the way to explicitly refresh diagnostics is better.

alxhub (approved):

alxhub (approved):

PR #1704, opened 2022-06-24
build: replace primary package manager yarn with pnpm

jbedard:


atscott:
The package manager we use is kept in line with other Angular repos. There are no plans to change that at this time.

jbedard:
This is the planned and required initial step toward transitioning to bazel and the [rules_js](https://github.com/aspect-build/rules_js/) ruleset. Or do we have to change our plans for that?

atscott:
@jbedard Ah, my apologies. I wasn't aware you were working with the DevRel team on this already. Could you add the whole context of why the change is being made to the commit message?

atscott:
// cc @devversion @josephperrott

devversion:
Agreed that commits, even if coordinated, should capture the context as much as possible.

I'm not aware of the concrete plans here, thinking we are still evaluating `rules_js` (also with its current RBE constraint). Generally I'm a huge fan of pnpm though 😄

gregmagolan:
Sorry for the lack of comms on this one. It was meant to be exploratory. We can stick with rules_nodejs + yarn on this repository if that is preferred.

Jason is trying out rules_js here as an example of using the new rule set (which relies on pnpm lockfile for lazy npm fetching & linking). The underlying switch between rules_nodejs or rules_js on this one should be trivial with rules_nodejs requiring rootDirs in the tsconfig and having more issues on Windows. It would be interesting to see if Windows development better with rules_js since there is no runtime linker to worry about. rules_js also doesn't work with RBE atm due to lack of symlink support for RBE but that is currently being worked on by Fabian Meumertzheim and may be fixed for Bazel 5.3.0.

devversion:
I'm personally up for giving this a shot here, but (1) Joey also needs to weight in and (2) how easy would it be to switch from `ts_project` to `ts_library`? We have plans to rework that rule to eventually also work with `rules_js` etc.

gregmagolan:
> I'm personally up for giving this a shot here, but (1) Joey also needs to weight in and (2) how easy would it be to switch from `ts_project` back to `ts_library`? It's just some extra manual tsconfig's right? We have plans to rework that rule to eventually also work with `rules_js` etc.

Probably not that bad to switch to ts_library either; @jbedard could make a commit to show what that could look like as well on top of this rules_js/ts_project PR.

devversion:
> Probably not that bad to switch to ts_library either; @jbedard could make a commit to show what that could look like as well on top of this rules_js/ts_project PR.
Not strictly necessary, if you can tell it's not likely too difficult. I can imagine that being the case. Certainly not opposed to seeing it concrete though.

gregmagolan:
We have a way to use rules_js now without having to switch to pnpm with
```
npm_translate_lock(
    name = "npm",
    yarn_lock = "//:yarn.lock",
    package_json = "//:package.json",
    verify_node_modules_ignored = "//:.bazelignore",
)
```

Closing this and I'll open up a new PR with the new method

PR #1723, opened 2022-07-09
feat(server): support code action

ivanwonder:
In ts service, the
`codeFixes`([getCodeFixesAtPosition](https://github.com/microsoft/TypeScript/blob/7584e6aad6b21f7334562bfd9d8c3c80aafed064/src/services/services.ts#L2689))
and
refactors([getApplicableRefactors](https://github.com/microsoft/TypeScript/blob/7584e6aad6b21f7334562bfd9d8c3c80aafed064/src/services/services.ts#L2699))
are resolved in different request. But in LSP they are resolved in the `CodeAction` request.
Now, this PR only handles the `codeFixes` because the `@angular/language-service` only supports it.

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -19,7 +19,7 @@ import {readNgCompletionData, tsCompletionEntryToLspCompletionItem} from './comp
 import {tsDiagnosticToLspDiagnostic} from './diagnostic';
 import {resolveAndRunNgcc} from './ngcc';
 import {ServerHost} from './server_host';
-import {filePathToUri, getMappedDefinitionInfo, isConfiguredProject, isDebugMode, lspPositionToTsPosition, lspRangeToTsPositions, MruTracker, tsDisplayPartsToText, tsTextSpanToLspRange, uriToFilePath} from './utils';
+import {filePathToUri, getMappedDefinitionInfo, isConfiguredProject, isDebugMode, lspPositionToTsPosition, lspRangeToTsPositions, MruTracker, tsDisplayPartsToText, tsFileTextChangesToLspWorkspaceEdit, tsTextSpanToLspRange, uriToFilePath} from './utils';
 
 export interface SessionOptions {
   host: ServerHost;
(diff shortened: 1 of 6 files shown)


dylhunn:
@ivanwonder Would you be able to rebase this PR? Let me know if it's time for review on this as well. Thanks!

ivanwonder:
> @ivanwonder Would you be able to rebase this PR? Let me know if it's time for review on this as well. Thanks!

@dylhunn Done. You can review it now.

server/src/session.ts
```diff
@@ -197,6 +200,59 @@ export class Session {
...
+        })
+        .concat(getCodeFixesAll(codeActions, params.textDocument));
+  }
+
+  private onCodeActionResolve(param: lsp.CodeAction): lsp.CodeAction {
+    const codeActionResolve = param.data as unknown as CodeActionResolveData;
+    const isCodeFixesAll = codeActionResolve.fixId !== undefined;
+    if (!isCodeFixesAll) {
+      return param;
+    }
```
atscott, line 238: Can you add a quick comment about why we don't need to do any resolving for things that aren't "codeFixesAll"? And a comment/link that describes why the presence or lack of a `fixId` is the distinguishing character for this type of fix.

server/src/session.ts
```diff
@@ -1239,3 +1299,34 @@ function isTypeScriptFile(path: string): boolean {
...
   return !isTypeScriptFile(path);
 }
+
+interface CodeActionResolveData {
+  fixId?: string;
+  document: lsp.TextDocumentIdentifier;
+}
+
+/**
+ * Extract the fixAll action from `codeActions`
```
atscott, line 1309: Similar to my comment in the `onCodeActionResolve` function about describing why we only need to resolve actions for "fixAll" code actions, can you add a little bit more to this documentation as well? That is, what exactly we're extracting and mention that this aren't "complete" code actions, but will need to be made "complete" in the `onCodeActionResolve` function.

atscott (approved):
LGTM, with a few nits and requests for additional comments. Great work!

dylhunn (approved):

atscott (approved):

PR #1733, opened 2022-08-04
build: make dist paths align with source file paths

gregmagolan:
This is a pre-factor for migrating to Bazel in https://github.com/angular/vscode-ng-language-service/pull/1732.

The `../common` requires are currently working because things are re-rooted in the `dist` folder and dist is a rootPath. Under bazel, output js files are in the same path structure as input ts files so relative requires should match the source tree structure since there is no re-rooting.


common/requests.ts
```diff
@@ -43,7 +43,7 @@ export const RunNgccRequest =
 export interface GetTcbResponse {
   uri: lsp.DocumentUri;
   content: string;
-  selections: lsp.Range[]
+  selections: lsp.Range[];
```
gregmagolan, line 46: lint

server/src/banner.ts
```diff
@@ -1,3 +1,11 @@
+/**
+ * @license
```
gregmagolan, line 2: lint

atscott (approved):

PR #1742, opened 2022-08-10
fix: support deeply nested pnpm virtual store node_modules paths in resolveAndRunNgcc

gregmagolan:
Uses indexOf instead of lastIndexOf in resolveAndRunNgcc.

This resolves the case where a user is using pnpm and ngcc is resolved through symlinks to a deeply nested virtual store node_modules location:

`some/user/path/node_modules/.pnpm/@angular/compiler-cli@1.2.3/node_modules/@angular/compiler-cli/...`

lastIndexOf would resolve to `some/user/path/node_modules/.pnpm/@angular/compiler-cli@1.2.3` which is incorrect. indexOf gets you to the desired `some/user/path`

### DIFF
--- a/server/src/ngcc.ts
+++ b/server/src/ngcc.ts
@@ -27,7 +27,7 @@ export async function resolveAndRunNgcc(tsconfig: string, progress: Progress): P
   if (!ngcc) {
     throw new Error(`Failed to resolve ngcc from ${directory}`);
   }
-  const index = ngcc.resolvedPath.lastIndexOf('node_modules');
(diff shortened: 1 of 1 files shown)


atscott (changes requested):
Sorry to be a stickler, but you you update the commit message to describe what's fixed and why it was broken rather than the "what" of the code change? This will help anyone going through the commits know why this change was made (otherwise, there's no context other than our conversation on slack)

atscott (approved):

PR #1747, opened 2022-08-27
fix(server): send diagnostic range to the Angular language service wh…

ivanwonder:
…en fixing code errors

This will make the fix error code simple, the range the fix error code get is always
the start and end of the diagnostic. For example, `<span>{{title}}</span>`, the `title` is
the missing member of component. When the range the user selects is from the start of `<span>`
to the end of `</span>`, the fix error code needs to check the start of the diagnostic.

### DIFF
--- a/integration/lsp/ivy_spec.ts
+++ b/integration/lsp/ivy_spec.ts
@@ -700,6 +700,33 @@ describe('code fixes', () => {
     expect(codeActions).toContain(jasmine.objectContaining(expectedCodeActionInTemplate));
   });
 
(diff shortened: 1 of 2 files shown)


dylhunn:
Sorry for the delay -- I was at ng-conf last week. Let me have a look.

dylhunn (approved):

dylhunn:
@ivanw

This has merge conflicts with the patch branch -- it will need another rebased PR if you want this to land in a patch version. But we're good for main.

PR #1793, opened 2022-10-28
fix(build): Adjust pointers to bazel-built directory

atscott:
Prior commits have migrated this project to have its outputs built by bazel. As a result the pointers to the dist folder need to be updated to bazel-bin.


client/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "rootDir": "src",
     "rootDirs": [
       ".",
-      "../dist"
+      "../dist",
+      "../bazel-bin"
```
gregmagolan, line 10: for editor support I'm guessing? or is this load bearing for debugging as well?

client/src/client.ts
```diff
@@ -490,7 +490,7 @@ function getServerOptions(ctx: vscode.ExtensionContext, debug: boolean): lsp.Nod
   // Node module for the language server
   const args = constructArgs(ctx, viewEngine, vscode.workspace.isTrusted);
   const prodBundle = ctx.asAbsolutePath('server');
-  const devBundle = ctx.asAbsolutePath(path.join('dist', 'server', 'src', 'server.js'));
+  const devBundle = ctx.asAbsolutePath(path.join('bazel-bin', 'server', 'src', 'server.js'));
```
gregmagolan, line 493: this one is surprising? needed for debugging or for some other reason?

package.json
```diff
@@ -198,16 +198,17 @@
     "onLanguage:html",
     "onLanguage:typescript"
   ],
-  "main": "./dist/client/src/extension",
+  "main": "./bazel-bin/client/src/extension",
```
gregmagolan, line 201: I think if you make this change you also have to update the substitution in the root BUILD file here:
https://github.com/angular/vscode-ng-language-service/blob/e808e621827f06fc68efe6fc1ba2f86fbfb70e3f/BUILD.bazel#L29

gregmagolan: Guessing "main" is load bearing on debugging somehow?

client/tsconfig.json
atscott: I'll revert this. I was seeing an error in my editor that was saying something along the lines of `...dist/common/resolver` wasn't built from source `...common/resolver.ts`. Which could make sense because we don't actually build anything with `tsc` and put it in the `dist` folder.

client/src/client.ts
atscott: This is when running the extension how locally. We don't copy anything from `bazel-bin` to `dist` so this is definitely necessary.

package.json
atscott: 👍 Done

atscott: > Guessing "main" is load bearing on debugging somehow?

More or less, yes I think. It's more generic https://docs.npmjs.com/cli/v8/configuring-npm/package-json#main but probably has to do with how the extension development host runs the extension

client/src/client.ts
gregmagolan: 👍

gregmagolan:
LGTM

devversion (approved):

dylhunn (approved):

PR #1794, opened 2022-10-28
build: update build.sh script for bazel

atscott:
There is no reason to copy files to the dist folder anymore with bazel. Update the build.sh to clean up previous builds by running 'bazel clean'.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -30,17 +30,9 @@ fi
 
 set -ex -o pipefail
 
-# Clean up from last build
-rm -rf dist
-
 # Run legacy yarn install to update the lock file since we may have modified the package.json above
 # and so that we can call bazel via yarn bazel
 yarn install
 
 # Build the npm package with bazel
 yarn bazel build //:npm --config=release
-
-# Copy the bazel built package to dist/npm
-mkdir -p dist/npm
-cp -r bazel-bin/npm/ dist/npm
-chmod -R +w dist/npm


gregmagolan:
Does the release process assume that output of the build is in dist?

atscott:
I already updated the wiki to refer to `bazel-bin` instead: https://github.com/angular/vscode-ng-language-service/wiki/Release

gregmagolan:
LGTM.

The files in bazel-bin are write-protected but shouldn't affect the release scripts since they won't be modifying them.

scripts/build.sh
```diff
@@ -31,16 +31,11 @@ fi
 set -ex -o pipefail
 
 # Clean up from last build
-rm -rf dist
+yarn bazel clean
```
devversion, line 34: Nit (just a statement): Generally I dislike cleaning before releasing as it avoids the benefits of cache and potentially is a symptom of bad build rules. I know that for stamping with `pkg_npm` this is historically needed though since the version is set as a volatile key instead of stable key (so that's a bug- had an issue for that somewhere).

(In Components we don't clean- but we just wipe the `pkg_npm` target to make sure it's re-stamped)

devversion (approved):

dylhunn (approved):

scripts/build.sh
atscott: Yea, I had added that because of the stamping that might result in multiple `vsix` files being in the `bazel-bin` (ones from previous builds and one from the current build since the `vsix` name included the version). I didn't want there to be any accident with copying/publishing the wrong version. 

Now that the build has removed the version from the `vsix` name, I think this can more safely be removed.

PR #1797, opened 2022-11-03
build: do not manage package version in Starlark file

devversion:
The package/project version should be managed via the `package.json` file so that existing tooling like `yarn ng-dev` does not break with the recent Bazel migration. Tools like `ng-dev` expect a project version to be present in the top-level `package.json`. The current placeholder breaks the release and branching expectations.

### DIFF
--- a/server/BUILD.bazel
+++ b/server/BUILD.bazel
@@ -1,53 +1,50 @@
 load("@aspect_rules_ts//ts:defs.bzl", "ts_config")
 load("@aspect_rules_esbuild//esbuild:defs.bzl", "esbuild")
 load("@aspect_rules_js//js/private:expand_template.bzl", "expand_template")
-load("//:version.bzl", "VERSION")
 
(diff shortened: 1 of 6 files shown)


server/BUILD.bazel
```diff
@@ -58,30 +55,32 @@ esbuild(
...
     name = "package_json_expanded",
-    template = "package.json",
     out = "package_expanded.json",
-    substitutions = {
-        "0.0.0-PLACEHOLDER": VERSION,
-    }
+    stamp_substitutions = {
+        "0.0.0-PLACEHOLDER": "{{BUILD_SCM_VERSION}}",
+    },
+    template = "package.json",
```
devversion, line 74: FYI: Mostly the `stamp_substitutions` are relevant for review. I kept the buildifier reformatting as they shouldn't hurt.

atscott (approved):

Issue #1815, opened 2022-11-18
Install Local @angular/language-service BROKEN under Bazel


gregmagolan:
Following the instructions at https://github.com/angular/vscode-ng-language-service/blob/main/DEVELOPER.md#install-local-angularlanguage-service no longer works with the Bazel build.

The way it used to work outside of Bazel is not compatible with a Bazel build as it references the npm package built by bazel in the angular/angular WORKSPACE by absolute path which is outside of the vscode-ng-language-service Bazel WORKSPACE and not a dependency that Bazel can consume.

The fix depends on some external changes:
- rules_js: https://github.com/aspect-build/rules_js/pull/621 & https://github.com/aspect-build/rules_js/pull/624
- angular/angular: https://github.com/angular/angular/pull/48120

PR #1816, opened 2022-11-18
fix: fix "install local @angular/language-service" workflow under Bazel build

gregmagolan:
Part of fix for #1815.

This is safe to land before any of the other fixes required do as it doesn't change anything here. Fix won't be in effect until rules_js & angular PRs land and this repo is updated to a release of rules_js that includes the its fixes.

See issue for details.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -56,6 +56,8 @@ build --enable_runfiles
 # Keep tests tagged "exclusive" in the sandbox
 test --incompatible_exclusive_test_sandboxed
 
+build --build_tests_only
+
 ###############################
 # Output                      #
 ###############################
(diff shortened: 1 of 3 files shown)


.bazelrc
```diff
@@ -56,6 +56,8 @@ build --enable_runfiles
 # Keep tests tagged "exclusive" in the sandbox
 test --incompatible_exclusive_test_sandboxed
 
+build --build_tests_only
```
gregmagolan, line 59: vsce complains in the release build when the local @angular/language-service packages are install. This flag makes it so `yarn test` only builds targets that are needed for the tests being run.

dylhunn (approved):

PR #1818, opened 2022-11-18
build: remove build-tests-only flag

dylhunn:
Greg asked me to remove this flag from the bazelrc: it controls what gets built when you use the ... selector, but it over filtered from what was intended.


atscott (approved):

atscott (changes requested):
Actually, I'm not sure this is relevant anymore? That line seems to have been updated to `test --build_tests_only` instead.

dylhunn:
Yeah, it's likely Greg fixed this out from under me :)

PR #1871, opened 2023-03-05
build: update to latest aspect_bazel_lib, aspect_rules_js, aspect_rules_jasmine and aspect_rules_esbuild

gregmagolan:
There is a regression with worker mode in this release of rules_ts for Node 14 (works fine with Node 16) I disabled it in the ts_project targets in the PR. Sahin is working on a fix for Node 14 worker mode but the ts_project targets in the repository are small so worker mode doesn't add much anyway.

The latest version of rules_js will keep your pnpm-lock.yaml files up-to-date automatically (when your source-of-truth is a yarn.lock file) when any of the inputs to the npm_translate_lock repository rule change. The `.aspect/rules/external_repository_action_cache/npm_translate_lock_*` files act like an action cache to know when to re-run `pnpm import` to update the pnpm lock files.

UPDATE: Node 14 worker mode issue is now fixed in rules_ts 1.3.2 (with rules_js 1.20.3 transitive dep)

PR #1873, opened 2023-03-07
Update aspect dependencies

devversion:


devversion:
This is a rebased/conflict resolved version of #1871. Since the v12 project got removed, the pnpm lock file had to be re-generated, and for some reason I cannot push to the Aspect org, even with "allow maintainer edits" enabled.

atscott (approved):

atscott:
I checked out this PR and verified that local debugging/breakpoints are still working.

PR #1875, opened 2023-03-09
Respect the client capability "textDocument.definition.linkSupport."

### LINKED ISSUE #1874: Always sending `LocationLink` and not respecting `textDocument.definition.linkSupport`
# 🐞 bug report

### Description

The LSP specification states that the client can request for additional meta data by specifying `textDocument.definition.linkSupport`, which I interpret as the client not supporting additional meta data if the property is not set.

This flag isn't respected by Angular Language Service as a `LocationLink` will always be send as a reply for `textDocument/definition` (go to definition), even to client which only supports `Location`

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#location  
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#locationLink

## Bug Type
The message `textDocument/definition` (go to definition) is not working on clients not supporting `LocationLink`.

- [x] Angular Language Service server

## Reproduction

Install a client which doesn't support `LocationLink`, and send the `textDocument/definition` message.

**Expected behavior**

Respect the specification.

Extension Version: 15.2.0

andlrc:
If the client doesn't explicitly tells the server that they have this capability then a `Location` should be send instead of `LocationLink` as a reply to a "Goto Definition Request".

See the relevant documentation:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition

Closes #1874

### DIFF
--- a/integration/lsp/test_utils.ts
+++ b/integration/lsp/test_utils.ts
@@ -73,6 +73,8 @@ export function initializeServer(client: MessageConnection): Promise<lsp.Initial
           }
         },
         moniker: {},
+        definition: {linkSupport: true},
+        typeDefinition: {linkSupport: true}
       }
     },
     /**
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -856,7 +856,8 @@ export class Session {
     return htmlLS.getFoldingRanges(virtualHtmlDoc);
   }
 
-  private onDefinition(params: lsp.TextDocumentPositionParams): lsp.LocationLink[]|null {
+  private onDefinition(params: lsp.TextDocumentPositionParams):
+      lsp.Location[]|lsp.LocationLink[]|null {
     const lsInfo = this.getLSAndScriptInfo(params.textDocument);
     if (lsInfo === null) {
       return null;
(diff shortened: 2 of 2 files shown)


integration/lsp/test_utils.ts
```diff
@@ -73,6 +73,7 @@ export function initializeServer(client: MessageConnection): Promise<lsp.Initial
           }
         },
         moniker: {},
+        definition: {linkSupport: true}
```
andlrc, line 76: I updated all existing tests to report as a client that supports `linkSupport`, as to not have the two existing "Go To Definition" tests failing.

server/src/session.ts
```diff
@@ -882,6 +892,14 @@ export class Session {
     if (!definitions) {
       return null;
     }
+
+    const clientSupportsLocationLinks =
+        this.clientCapabilities.textDocument?.definition?.linkSupport ?? false;
```
atscott, line 897:
```suggestion
        this.clientCapabilities.textDocument?.typeDefinition?.linkSupport ?? false;
```

nit: I think technically this should be `typeDefinition` though I doubt these would ever truly be different in the client capabilities. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_typeDefinition

integration/lsp/test_utils.ts
atscott: 👍

atscott (approved):
Looks good other than 1 technicality in `onTypeDefinition`. Thanks for the fix!

andlrc:
> Looks good other than 1 technicality in `onTypeDefinition`. Thanks for the fix!

Thanks for taking the time to guide me and review the code.

The main concern that I could have with this PR is that we need to ensure that the  language server client running inside vscode (or any other consumer) announces their client capabilities correctly, as in including `linkSupport` in the object, or gracefully handles both `Location` and `LocationLink`.

I'm not sure how this can be tested in vscode, or if the automatic tests already covers this.

atscott:
> The main concern that I could have with this PR is that we need to ensure that the language server client running inside vscode (or any other consumer) announces their client capabilities correctly, as in including linkSupport in the object, or gracefully handles both Location and LocationLink.

I checked out this PR and double checked that it's working in VSCode and that the VSCode client correctly sends its capabilities.

> I'm not sure how this can be tested in vscode, or if the automatic tests already covers this.

Unfortunately the integration tests which would actually launch the VSCode client with a local build have been disabled due to flakiness. Some day I hope to resurrect them but at the moment, tests which need to exercise end-to-end behavior like interactions between client and server are manual.

atscott (approved):

Issue #1902, opened 2023-07-13
JSDoc of component properties or events in template is not displayed properly


KingMario:
# 🐞 bug report

### Is this a regression?

No

### Description

JSDoc information of a component property or event should be displayed in template. But if the JSDoc is long or have a blank leading line, or have block tags such as `@description`, the information is not displayed properly. The JSDoc information can be displayed properly in `.ts` file.

For example, given the following property with JSDoc, in `.ts` file, the whole information is displayed when the property is hover over, but in HTML template, it only displays `*`.

```javascript
  /***
   * @description
   * whether to filter null value if no item is selected
   ***/
  @Input() nullValueFilterable = false;
```

Expected is all the information should be displayed in HTML template.

## Bug Type
What does this bug affect

- [x] Angular Language Service VSCode extension
- [ ] Angular Language Service server

### **Screenshots**

JSDoc information displayed in `.ts` file:

<img width="618" alt="image" src="https://github.com/angular/vscode-ng-language-service/assets/3956876/e00c8074-04a6-4d77-ab65-fcd57146c0a8">

`*` displayed in HTML template:

<img width="578" alt="image" src="https://github.com/angular/vscode-ng-language-service/assets/3956876/31156e2d-4434-4e2b-92e0-9c0105c105e2">

It seems that, in HTML template, the JSDoc is displayed if ONLY the JSDoc is in one line and has no block tags such as `@desc`, `@example` or so.

For example, after update the JSDoc to the following text without the `@description` block tag, the information is displayed in HTML template (but treated as dotted list with a trailing blank line).

```
  /***
   * whether to filter null value if no item is selected
   ***/
```

<img width="575" alt="image" src="https://github.com/angular/vscode-ng-language-service/assets/3956876/f9f6063f-82e8-4c17-9b30-d2895baaca7a">

## 🌍  Your Environment
**Angular Version:**
<pre><code>
     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 14.2.11
Node: 16.19.1
Package Manager: npm 9.7.2 
OS: darwin x64

[...]

PR #1904, opened 2023-07-16
fix: support to show the tag info in the jsDoc

ivanwonder:
https://github.com/microsoft/vscode/blob/18136ed6ec4bfabc58664c8b656839709ea997e8/extensions/typescript-language-features/src/languageFeatures/util/textRendering.ts

https://github.com/microsoft/vscode/blob/18136ed6ec4bfabc58664c8b656839709ea997e8/extensions/typescript-language-features/src/test/unit/textRendering.test.ts

The code is copied from the VSCode extension. The difference is that the code runs in the server in the Angular extension, and runs in the client in the VSCode extension

Fixes https://github.com/angular/vscode-ng-language-service/issues/1902

cc @dylhunn

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -59,6 +59,9 @@ export class AngularLanguageClient implements vscode.Disposable {
       // Don't let our output console pop open
       revealOutputChannelOn: lsp.RevealOutputChannelOn.Never,
       outputChannel: this.outputChannel,
+      markdown: {
+        isTrusted: true,
+      },
       middleware: {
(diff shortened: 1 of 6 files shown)


atscott:
I would have preferred not needing to do the jsdoc=>markdown conversion in the repo but since I don't have another suggestion, I can't defend blocking this bug fix. Maybe it can be updated at some point in the future.

atscott (approved):
Thank you for the fix!

PR #1937, opened 2023-10-09
feat: support block completions

dylhunn:
Two changes are necessary to support autocompletion of block keywords:
1. The `@` keyword causes immediate completion
2. The new block completion item kind maps onto the keyword icon


dylhunn:
(Related to angular/angular#52121)

atscott (approved):

PR #1953, opened 2023-11-03
Syntax cleanup

atscott:
Cleans up some syntax configurations. See individual commits.

### DIFF
--- a/syntaxes/inline-template.json
+++ b/syntaxes/inline-template.json
@@ -69,6 +69,15 @@
         },
         {
           "include": "template.ng"
+        },
+        {
+          "include": "template.blocks.ng"
+        },
+        {
+          "include": "template.tag.ng"
+        },
+        {
+          "include": "template.let.ng"
         }
       ]
     }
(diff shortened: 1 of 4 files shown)


syntaxes/inline-template.json
```diff
@@ -69,6 +69,9 @@
         },
         {
           "include": "template.ng"
+        },
+        {
+          "include": "template.blocks.ng"
```
dylhunn, line 74: How were the blocks syntaxes previously being included, if this line is necessary to add them to the main template syntaxes?

dylhunn (approved):

syntaxes/inline-template.json
atscott: Blocks were included in TS files via the `injectTo: "source.ts"` before. Which actually means that the block syntax highlighting would apply to _any_ code in a TS file that switched to the HTML language, not just strings assigned to `template`.

This change makes the inclusion more specific and targeted -- we only add the block syntax pattern under strings that we have identified as inline templates.

PR #1967, opened 2023-11-09
refactor(syntaxes): Update embedded languages and block name

atscott:
This commit updates the embedded languages of template.blocks.ng to more accurately reflect content. It also makes expressions switch the javascript language. Lastly, this updates the block capture to name the entire content control.block.ng rather than just contentName, which excluded the block name and expression.


integration/lsp/ivy_spec.ts
```diff
@@ -580,6 +580,56 @@ export class AppComponent {
     });
     expect(componentResponse).toBe(true);
   })
+
+  describe('auto-import component', () => {
```
dylhunn, line 584: Is it a rebase mistake that so much auto-import and quick-action code is included in this PR?

atscott: Rebase mistake!

dylhunn: Thanks! Still wrapping my head around the grammar change. Will ping you later today with thoughts and maybe to chat

dylhunn (approved):
I think this makes sense, and the updated tests look good (although I'm admittedly not 100% on it).

PR #1992, opened 2024-01-30
Remove eventBinding and propertyBinding

### LINKED ISSUE #1725: Binding syntax outside of an attribute still highlighted as a binding rather than plain text
# 🐞 bug report

### Is this a regression?

N/A

### Description

 A clear and concise description of the problem...

If the bug is caused or experienced by a particular source code, please provide a minimal
reproduction of that source code.

## Bug Type
What does this bug affect

-  [x] Angular Language Service VSCode extension

- [ ] Angular Language Service server

## Reproduction

1. in a component.html file add the following:
```
    <app-image-img>
        [filesArray]="event.Event_Files" //this line of text is blue!
    </app-image-img>
```

**Expected behavior**

- Because the ```filesArray``` property is defined outside the closing arrow of the opening ```<app-image-img>``` element, I would expect the text to turn white as it does when a property without an array is added outside the opening arrow:

```
    <app-image-img>
        [filesArray]="event.Event_Files" //this line of  text is blue, indicating no error!
        files="test" //this line of text is white, denoting an error!
    </app-image-img>
```

### **Logs**

N/A - No errors in Angular Server Log or Browser Console

### **Screenshots**

Incorrectly added property tag outside of closing arrow: (unexpected behavior, text is blue.)
image

Shows white text when non-bracketed property is added: (expected behavior, showing white text)
image

Correct implementation of property binding: (working properly)
image

## 🌍  Your Environment
**Angular Version:**
<pre><code>
Angular CLI: 13.3.7
Node: 16.13.2
Package Manager: npm 8.1.2
OS: win32 x64
</code></pre>

Extension Version:
<pre><code>
Angular Language Service v14.01
</code></pre>

VSCode Version:
<pre><code>
Version: 16.9.0 (user setup)
</code></pre>

Operating System:
<pre><code>
Edition	Windows 10 Pro
Version	21H2
Installed on	‎5/‎28/‎2021
OS build	19044.1766
Experience	Windows Feature Experience Pack 120.2212.4180.0
</code></pre>

Extension options:
<pre><code>
{
    "angular.log": "verbose"
}
</code></pre>

**Anything else relevant?**
N/A

a-stewart:
This removes the Angular definitions of event and property bindings, in favour of those provided by the default HTML extension in VS Code.

This fixes #1725

See https://github.com/angular/vscode-ng-language-service/issues/1725#issuecomment-1671803141 for justification

### DIFF
--- a/syntaxes/src/template.ts
+++ b/syntaxes/src/template.ts
@@ -13,8 +13,6 @@ export const Template: GrammarDefinition = {
   injectionSelector: 'L:text.html -comment',
   patterns: [
     {include: '#interpolation'},
-    {include: '#propertyBinding'},
-    {include: '#eventBinding'},
     {include: '#twoWayBinding'},
     {include: '#templateBinding'},
   ],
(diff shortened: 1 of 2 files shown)


atscott:
@a-stewart Hey, sorry I hadn't pinged you on the issue report. I've been working with the syntaxes a bit more recently to support the control flow and realized that removing these isn't a very good option. This would remove the expression highlighting on the value side. On the key side, we don't _really_ need any special highlighting, but the value side includes the `expression.ng` which includes a lot more Angular-specific template syntax highlighting.

a-stewart:
No worries

Another approach which could be taken would be to have another file, eg, template-tags.json, which we inject into `soure.html meta.tag` or something like that.

atscott:
> Another approach which could be taken would be to have another file, eg, template-tags.json, which we inject into soure.html meta.tag or something like that.

Yea, that's probably the best path forward. We'd have more control there to include/exclude the syntaxes in certain contexts.

a-stewart:
Let me have a quick play and then I'll get back to you if it seems to work.

a-stewart:
I need to call it a day now as it is late here, but #1993 (WIP) seems to work as expected.

PR #2027, opened 2024-04-27
refactor(client): use `SourceFile` to detect the Angular context in t…

ivanwonder:
…he client

When using the typescript's scanner to parse the template literals which includes an expression, knowing the right `CloseBraceToken` of the expression to invoke the `reScanTemplateToken` function is a little complex. If don't do this, the rest of the file is temporarily "poisoned".

There is an explanation for why typescript's fast-acting lexical classifications don't work for template sting across lines.
https://github.com/microsoft/TypeScript/issues/1477#issuecomment-66907946 https://github.com/microsoft/TypeScript/blob/a4d12a46c8b413c65a730c4ad0323459f1fc44ce/src/services/classifier.ts#L114

So this PR uses the `SourceFile` to detect the Angular context in the client.

### DIFF
--- a/client/src/embedded_support.ts
+++ b/client/src/embedded_support.ts
@@ -14,8 +14,13 @@ export function isInsideInlineTemplateRegion(
   if (document.languageId !== 'typescript') {
     return true;
   }
-  return isPropertyAssignmentToStringOrStringInArray(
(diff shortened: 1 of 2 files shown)


atscott (approved):
Great, thanks for doing this!

PR #2042, opened 2024-06-24
feat(syntaxes): add support for let declarations

crisbeto:
Adds highlighting support for the new `@let` syntax.

Example inside vscode:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -186,6 +186,14 @@
         }
       },
       {
+        "path": "./syntaxes/let-declaration.json",
+        "scopeName": "template.let.ng",
+        "injectTo": [
+          "text.html.derivative",
+          "source.ts"
+        ]
+      },
+      {
         "path": "./syntaxes/template-tag.json",
         "scopeName": "template.tag.ng",
         "injectTo": [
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -185,6 +185,18 @@
...
+      {
+        "path": "./syntaxes/let-declaration.json",
+        "scopeName": "template.let.ng",
+        "injectTo": [
+          "text.html.derivative",
+          "source.ts"
+        ],
+        "embeddedLanguages": {
+          "text.html": "html",
+          "expression.ng": "javascript"
```
atscott, line 197: I'm pretty sure these can be removed. I have a PR that would clean this up for some of the other scopes (https://github.com/angular/vscode-ng-language-service/pull/1953/commits/066c1d63a445332d34fb96a369d1c069443b338a). Since none of the scopes in the definition have the `text.html` or `expression.ng` names, I don't think this is used.

crisbeto: Done.

crisbeto (commented):
Feedback is addressed. Also fixed the CI.

atscott (approved):

dylhunn (approved):

PR #2050, opened 2024-07-08
feat: integrate code refactorings from language-service via LSP

devversion:
This integrates Angular's code refactorings into the extension, so that those are communicated via the LSP to VSCode.

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -66,9 +66,8 @@ export class AngularLanguageClient implements vscode.Disposable {
         provideCodeActions: async (
             document: vscode.TextDocument, range: vscode.Range, context: vscode.CodeActionContext,
             token: vscode.CancellationToken, next: lsp.ProvideCodeActionsSignature) => {
-          if (await this.isInAngularProject(document) &&
-              isNotTypescriptOrInsideComponentDecorator(document, range.start) &&
-              isNotTypescriptOrInsideComponentDecorator(document, range.end)) {
+          // Code actions can trigger also outside of `@Component(<...>)` fields.
+          if (await this.isInAngularProject(document)) {
             return next(document, range, context, token);
           }
         },
--- a/.aspect/rules/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU=
+++ b/.aspect/rules/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU=
@@ -2,7 +2,7 @@
 # Input hashes for repository rule npm_translate_lock(name = "npm", pnpm_lock = "//:pnpm-lock.yaml").
 # This file should be checked into version control along with the pnpm-lock.yaml file.
 .npmrc=974837034
-pnpm-lock.yaml=-1342496482
-yarn.lock=-1243860663
-package.json=1843460007
+pnpm-lock.yaml=-1152034998
+yarn.lock=-1189064062
+package.json=-1838831543
 pnpm-workspace.yaml=1711114604
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -37,7 +37,7 @@
     {
       "type": "shell",
       "label": "build vsix",
-      "command": "./scripts/build.sh"
+      "command": "./scripts/build.sh package.json"
     }
   ]
 }
--- a/package.json
+++ b/package.json
@@ -231,7 +231,7 @@
     "test:legacy-syntaxes": "yarn compile:syntaxes-test && yarn build:syntaxes && jasmine dist/syntaxes/test/driver.js"
   },
   "dependencies": {
-    "@angular/language-service": "18.1.0-rc.0",
+    "@angular/language-service": "~18.2.0-next.1",
     "typescript": "5.5.3",
     "vscode-html-languageservice": "^4.2.5",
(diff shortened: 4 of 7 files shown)


devversion:
Blocked on https://github.com/angular/angular/pull/56895

client/src/client.ts
```diff
@@ -66,9 +66,7 @@ export class AngularLanguageClient implements vscode.Disposable {
         provideCodeActions: async (
             document: vscode.TextDocument, range: vscode.Range, context: vscode.CodeActionContext,
             token: vscode.CancellationToken, next: lsp.ProvideCodeActionsSignature) => {
-          if (await this.isInAngularProject(document) &&
-              isInsideInlineTemplateRegion(document, range.start) &&
-              isInsideInlineTemplateRegion(document, range.end)) {
+          if (await this.isInAngularProject(document)) {
```
atscott, line 69: Hmm, we should keep an eye on reports of performance regressions with this change. I don't know if this particular request depends greatly on quick turnaround like completions would. 

`onCodeAction` does call `getCodeFixesAtPosition` which triggers semantic diagnostics (and I think global compilation) so it could be pretty slow in a large project/file. Maybe we could optimize this by adding an endpoint to the language service that can first check if there's a `CodeActionMeta` for the diagnostic error code and only then ask for the code fixes.

atscott (approved):

client/src/client.ts
devversion: I think this is reasonable. Let me look into this.  

Though I wonder if the performance problem you were addressing back then, actually was addressed by this. Based on the issue conversation it didn't seem like it?

ivanwonder: When I call the `getSemanticDiagnostics` in the code action, I think it should not have a performance issue.

1. This only triggers the currently active file semantic diagnostics. The time spent should be the same as the completion.

2. The vscode should trigger the code action after the diagnostics changes. So When the code action is triggered, the semantic diagnostics have been computed. the second time triggering the diagnostics should hit the cache(If not, I think the code action can be canceled, and wait for the vscode to trigger the semantic diagnostics first). We can cache the diagnostics in the `program`(or `SourceFile`) or the `NgCompiler` has done it?

Correct me if I'm wrong.

devversion: I'm not too familiar with this code in concrete here, but couldn't VSCode trigger a code action on a file that only has semantic diagnostics checked for TypeScript— but not the Angular extension?

I assume, probably not the case, but in either case, it wouldn't hurt having a safety quick exit just to be sure?

atscott: > This only triggers the currently active file semantic diagnostics. The time spent should be the same as the completion.

Yes, but this actually still requires the whole program to be analyzed:

https://github.com/angular/angular/blob/982f1b125147e4292716f9524bef75423b70c71c/packages/compiler-cli/src/ngtsc/core/src/compiler.ts#L596
https://github.com/angular/angular/blob/982f1b125147e4292716f9524bef75423b70c71c/packages/compiler-cli/src/ngtsc/core/src/compiler.ts#L1156
https://github.com/angular/angular/blob/982f1b125147e4292716f9524bef75423b70c71c/packages/compiler-cli/src/ngtsc/core/src/compiler.ts#L930-L951

ivanwonder: > I assume, probably not the case, and what you write makes sense to me— but in either case, it wouldn't hurt having a safety quick exit just to be sure?

No, it wouldn't hurt to have this. I am just curious about what caused this performance issue.

> I'm not too familiar with this code in concrete here, but couldn't VSCode trigger a code action on a file that only has semantic diagnostics checked for TypeScript— but not the Angular extension?

The code action will be skipped because it's out of the angular context.

> Yes, but this actually still requires the whole program to be analyzed:

But the completion also requires the whole program to be analyzed, because it depends on the template type checker:

https://github.com/angular/angular/blob/d97aee6dc2266c81cbd5207f642f3fee39a2d414/packages/language-service/src/completions.ts#L124

https://github.com/angular/angular/blob/d97aee6dc2266c81cbd5207f642f3fee39a2d414/packages/compiler-cli/src/ngtsc/core/src/compiler.ts#L701

atscott: > But the completion also requires the whole program to be analyzed, because it depends on the template type checker

Yes! Pretty much everything requires it and that’s why we try to avoid making these requests in places where we don’t have to. When we didn’t have these early-returns, we had performance issues files in the repo, especially for completions, where we were slowing down the completion results for every typescript file, even outside the inline template where we’d have no additional completion information.

ivanwonder: Thanks, it makes sense to me.

ivanwonder:
@atscott Take a look at this [PR](https://github.com/angular/angular/pull/57165).

A public API called [getSupportedCodeFixes](https://github.com/microsoft/TypeScript/blob/6d3be985c82bead3b41348de76efec8110c677c5/src/services/types.ts#L697) is used to check if the compiler can fix the diagnostics. Remove the unnecessary export custom API.

devversion:
Note: This PR also bumps to a next release of the language service. Not sure if this is fine for the versioning in this repository— but I assume yes, given that there is a 18.x branch

atscott (approved):

PR #2067, opened 2024-07-16
efix(extension): Allow requests to reach the language server when in c…

### LINKED ISSUE #2064: quick info stops working if there are two components in the same file and I change a field in parent component
# 🐞 bug report

### Description

https://github.com/angular/vscode-ng-language-service/assets/25394362/a30843f3-e7c5-4106-b8a3-4c676389963e

## Bug Type
What does this bug affect

- [x] Angular Language Service VSCode extension
- [x] Angular Language Service server

**Expected behavior**

Quick info to not stop working

atscott:
…omponent decorator

This commit updates the early-return logic on the client-side to allow requests to go to the server whenever the cursor is inside the component decorator. Prior to this change, some requests would attempt to return early if the position was not inside the inline template region. However, we were never able to get the tokenizing working correctly and this has led to issues whenever there are multiple template strings (JS template strings with \`).

This approach should give us the needed benefits of early-returns in non-Angular contexts (`@Component` decorator should be pretty rare) while being more permissive and correct.

fixes #2064

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -15,7 +15,7 @@ import {OpenOutputChannel, ProjectLoadingFinish, ProjectLoadingStart, SuggestStr
 import {GetComponentsWithTemplateFile, GetTcbRequest, GetTemplateLocationForComponent, IsInAngularProject} from '../../common/requests';
 import {NodeModule, resolve} from '../../common/resolver';
 
-import {isInsideComponentDecorator, isInsideInlineTemplateRegion, isInsideStringLiteral} from './embedded_support';
+import {isInsideStringLiteral, isNotTypescriptOrInsideComponentDecorator} from './embedded_support';
 
 interface GetTcbResponse {
   uri: vscode.Uri;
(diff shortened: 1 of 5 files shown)


client/src/client.ts
```diff
@@ -67,8 +67,8 @@ export class AngularLanguageClient implements vscode.Disposable {
             document: vscode.TextDocument, range: vscode.Range, context: vscode.CodeActionContext,
             token: vscode.CancellationToken, next: lsp.ProvideCodeActionsSignature) => {
           if (await this.isInAngularProject(document) &&
-              isInsideInlineTemplateRegion(document, range.start) &&
-              isInsideInlineTemplateRegion(document, range.end)) {
+              isInsideComponentDecorator(document, range.start) &&
+              isInsideComponentDecorator(document, range.end)) {
```
devversion, line 71: The function seems to also return for HTML files, but the name only suggests "inside decorator". Wondering if we can rename, but otherwise LGTM.

devversion (approved):

PR #2106, opened 2024-11-08
fix: improve detection of Angular core version in monorepo setup

devversion:
Angular may not be installed at the project root. We should look for `package.json` roots and determine the `@angular/core` version based on these folders.

Related to: https://github.com/angular/angular/issues/58546

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -227,7 +227,7 @@ export class AngularLanguageClient implements vscode.Disposable {
     }
 
     // Node module for the language server
-    const args = this.constructArgs();
+    const args = await this.constructArgs();
     const prodBundle = this.context.asAbsolutePath('server');
     const devBundle =
         this.context.asAbsolutePath(path.join('bazel-bin', 'server', 'src', 'server.js'));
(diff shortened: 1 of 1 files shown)


client/src/client.ts
```diff
@@ -554,11 +554,15 @@ function extensionVersionCompatibleWithAllProjects(serverModuleLocation: string)
 /**
  * Returns true if any project in the workspace supports block syntax (v17+).
```
crisbeto, line 555: nit: this dec string is quite outdate by now.

client/src/client.ts
```diff
@@ -554,11 +554,15 @@ function extensionVersionCompatibleWithAllProjects(serverModuleLocation: string)
...
  */
-function getAngularVersionsInWorkspace(): NodeModule[] {
+async function getAngularVersionsInWorkspace(outputChannel: vscode.OutputChannel): Promise<NodeModule[]> {
+  const packageJsonFiles = await vscode.workspace.findFiles("**/package.json", "**/node_modules/**");
```
crisbeto, line 558: Should we change the `node_modules` pattern to be something like `**/node_modules/@angular/core/**` to narrow down the amount of directories it needs to traverse?

crisbeto (approved):

client/src/client.ts
devversion: This is just the exclusion pattern. We'd want to exclude all `node_module` folders completely. I will add a comment.

Issue #2119, opened 2024-11-20
The LS does not recognise components as being standalone


JulienLecoq:
# 🐞 bug report

### Description

Since ``Angular 19``, components are standalone by default. However, the LS does not seems to take that into account. I've migrated my Angular app to Angular 19 so now all my components does not have anymore the ``standalone: true`` line but this cause errors to appears in every template and in every imports: [] because the LS does not consider my components to be standalone.

ps: I don't use VSCode but I use Neovim.

## Bug Type
What does this bug affect

- [ ] Angular Language Service VSCode extension
- [x] Angular Language Service server

## Reproduction

1. Create a new Angular project
2. Migrate the project to Angular 19
3. See the error in app.component.ts saying that 'imports' is only valid on a component that is standalone.

**Expected behavior**

No error should appear.

### **Screenshots**

![Image](https://github.com/user-attachments/assets/6b121391-0298-4075-b577-2b7c08029792)

![Image](https://github.com/user-attachments/assets/0dfcb97b-6faf-4a4f-b936-0d1583b71ff8)

## 🌍  Your Environment
**Angular Version:**

<pre><code>
Angular CLI: 19.0.0
Node: 20.14.0
Package Manager: npm 10.9.0
OS: darwin arm64

Angular: 19.0.0
... animations, cli, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1900.0
@angular-devkit/build-angular   19.0.0
@angular-devkit/core            19.0.0
@angular-devkit/schematics      19.0.0
@schematics/angular             19.0.0
rxjs                            7.8.1
typescript                      5.5.4
zone.js                         0.15.0
</code></pre>

Extension Version:
<pre><code>
v19.0.2
</code></pre>

Operating System:
<pre><code>
macOS Sequoia : 15.0.1
</code></pre>

mrentmeister-tt:
I am experiencing this issue as well after updating my Nx monorepo to v19. VSCode extension version 19.0.2.

JulienLecoq:
@mrentmeister-tt Mhh, did you try to downgrade the version to v19.0.0 if you can with VSCode? I'm using Mason which automatically manage the language server version so I can't do it with it. I'll try to do it manually without Mason tomorrow to see if it resolve the issue.

HymanZHAN:
[...]

PR #2121, opened 2024-11-27
build: update to the latest Angular version

crisbeto:
Updates to the latest version of Angular which includes some language service fixes.


devversion (approved):

dzonatan:
Can this be released? It should fix the #2119

Issue #2137, opened 2025-01-20
bug: extra ] gets added on attribute binding completion


eneajaho:
# 🐞 bug report

### Is this a regression?

No

### Description

 A clear and concise description of the problem...

<p [title|]></p> 

- Press enter 

Produces -> <p [title]=""]></p> 

## Reproduction

https://github.com/user-attachments/assets/f3879aff-e20a-48fe-b8da-d4fe118b931c

eneajaho:
It happens for events too

ivanwonder:
@eneajaho Setting `editor.suggest.insertMode` to `Replace` can fix the problem, but this is not the correct approach.

The Angular extension should prevent developers from altering this behavior. but `InsertReplaceEdit` provides developers with options. Therefore, we can remove the [branch](https://github.com/angular/vscode-ng-language-service/blob/5d67255385e31ab47f01a38fd28cedfe3e0ebcd9/server/src/completion.ts#L147) and use `TextEdit` instead.

I don't have time to fix this now, but I think it's easy.

Maybe @atscott can take a moment to look at this. 

https://code.visualstudio.com/updates/v1_51#_status-bar-for-suggestions
https://github.com/microsoft/vscode/issues/10266#issuecomment-555569954

PR #2142, opened 2025-02-09
fix: replace the closing symbol for inputs/outputs.

ivanwonder:
The Angular Language Service does not return `InsertReplaceEdit`. There is no need to allow the developer to choose how to insert the completion.

```
For example, `<button (c|) />`.
                      ^^__________Insert edit
                      ^^ ^________Replace edit
```
If the LS returns the `InsertReplaceEdit` as shown above, selecting "Insert" by the developer results in `(click)="")`, and selecting "Replace" results in `(click)=""`.

Now in the vscode, the default `editor.suggest.insertMode` value for HTML is `Replace`, for ts is `Insert`, So this leads to a bug in the ts file.

Fixes https://github.com/angular/vscode-ng-language-service/issues/2137

### DIFF
--- a/server/src/completion.ts
+++ b/server/src/completion.ts
@@ -106,8 +106,8 @@ function ngCompletionKindToLspCompletionItemKind(kind: CompletionKind): lsp.Comp
  * @param scriptInfo
  */
 export function tsCompletionEntryToLspCompletionItem(
-    entry: ts.CompletionEntry, position: lsp.Position, scriptInfo: ts.server.ScriptInfo,
-    insertReplaceSupport: boolean): lsp.CompletionItem {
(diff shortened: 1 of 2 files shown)


ivanwonder:
This also reverts the [commit](https://github.com/angular/vscode-ng-language-service/commit/c4310e2835c3506dcf73f277bdd929ae32c145d9#diff-b4f011ffc815aeebd238be700717f41350c9c094198e94d18e358c73580f5d2eL117).

PTAL @atscott

atscott (approved):
LGTM with the caveat that this applies to all completions, not just inputs and outputs with their parens/braces. This effectively removes the support for `insert` mode and uses replace everywhere.

Issue #2153, opened 2025-03-12
Error compiling in Visual Studio Code - Error with angular/core?


jsgoupil:
# 🐞 bug report

### Is this a regression?

Yes, not sure which version exactly

### Description

 A clear and concise description of the problem...

## Bug Type
What does this bug affect

- [X] Angular Language Service VSCode extension
- [ ] Angular Language Service server

## Reproduction

1. `ng new test`
2. Open app.component.ts
3. add `changeDetection: ChangeDetectionStrategy.Default,`

Does not compile well in Visual Studio Code.
```
[{
	"resource": "/f:/git/clients/test/src/app/app.component.ts",
	"owner": "_generated_diagnostic_collection_name_#4",
	"code": "-991010",
	"severity": 8,
	"message": "changeDetection must be a member of ChangeDetectionStrategy enum from @angular/core\n  Value is of type 'ChangeDetectionStrategy$1'.",
	"startLineNumber": 7,
	"startColumn": 20,
	"endLineNumber": 7,
	"endColumn": 51
}]
```

**Expected behavior**

Suppose to compile properly.

### **Logs**

<pre><code>
</code></pre>

### **Screenshots**

![Image](https://github.com/user-attachments/assets/b6ef0acb-da5b-4973-ba84-6b8c68813c7e)

## 🌍  Your Environment
**Angular Version:**
<pre><code>

Angular CLI: 19.2.2
Node: 20.11.1
Package Manager: npm 8.5.5
OS: win32 x64

Angular: 19.2.2
... cli, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1902.2
@angular-devkit/build-angular   19.2.2
@angular-devkit/core            19.2.2
@angular-devkit/schematics      19.2.2
@schematics/angular             19.2.2
rxjs                            7.8.2
typescript                      5.7.3
zone.js                         0.15.0

</code></pre>

Extension Version:
<pre><code>

19.2

</code></pre>

VSCode Version:
<pre><code>

Version: 1.98.1

</code></pre>

Operating System:
<pre><code>

Windows 11 Pro
Version	10.0.26100 Build 26100

</code></pre>

Extension options:
<pre><code>

</code></pre>

**Anything else relevant?**
Brand new project.

atscott:
[...]

PR #2154, opened 2025-03-13
build: update language-service version and bump version for release

devversion:
Updates the language-service version to fix issues like: https://github.com/angular/vscode-ng-language-service/issues/2153

Bumps the version of the repository so that we can cut a new VSCode extension release.

I've verified locally that this fixes the issues as expected. We just didn't have a extension release with the latest version.


crisbeto (approved):

devversion:
Verified locally that all tests pass (because CI is probably still OOM-ing currently)

PR #2155, opened 2025-03-14
feat: add support for host bindings

crisbeto:
Makes the necessary changes on the vscode extension side to enable https://github.com/angular/angular/pull/60267, including:
* Looking for completions, quick info etc inside the `host` property.
* Syntax highlighting for the `host` object literal.


atscott (approved):

crisbeto:
I had to change the approach here a bit, because the previous one wasn't accounting for a bunch of TS syntax like spread expressions or escaped strings. The new approach will fall back to the TS highlighting if our patterns don't match anything and it should be a bit more robust.

PR #2161, opened 2025-03-20
fix(syntaxes): `keyword` token for `track`

dannymcgee:
This PR updates the TM grammar for flow-control block expressions to tokenize `track` as a keyword.

### Changes made
- Separate the valid sub-expressions of a `@for`/`@if` header into their own "clause" patterns (helps to avoid highlighting `track` as a keyword in the wrong context)
- Assign `keyword.control.track.ng` scope to the `track` token at the beginning of a "track-by" clause
- The `item` binding in `item of items` is now `constant` instead of `readwrite`, following the same logic mentioned in #2160. Again, happy to revert this change if that's unwanted.

#### Before:
track-highlighting-before

#### After:
track-highlighting-after


atscott (approved):
Nice, thanks for this

PR #2165, opened 2025-03-28
fix(language-server): Do best-effort to initialize composite project …

### LINKED ISSUE #2149: Go to Definition from angular template no longer working
# 🐞 bug report

### Is this a regression?

YES functioned in 19.0.4. Broken in 19.1 and 19.2

### Description

When in an angular html template neither right click -> go to definition or f12 or cmd+click are going to definition. 

## Bug Type
- [x] Angular Language Service VSCode extension
- [ ] Angular Language Service server

## Reproduction

1. Have the latest version of Angular Language Service installed in vs code
2. go to an angular html template
3. try to go to definition on a function that is being used in the template

**Expected behavior**

User should be able to go to definition.

## 🌍  Your Environment
**Angular Version:**
<pre><code>
19.0.6
</code></pre>

Extension Version:
<pre><code>
19.2.0
</code></pre>

VSCode Version:
<pre><code>
1.97.2 (Universal)
</code></pre>

Operating System:
<pre><code>
MacOS Sonoma 14.3
</code></pre>

Extension options:
<pre><code>
defaults
</code></pre>

**Anything else relevant?**
This was reproducible by every person on my team (4 of us). Reverting to 19.0.4 fixed it for all of us.

atscott:
…when HTML is opened first

An upstream TypeScript performance-motivated change ( https://github.com/microsoft/TypeScript/commit/02b07a1788c8ec5b47e51b11955e72ee6d175f4b) made it so composite projects are not opened unless the opened file is in the includes/files list of the `tsconfig`. This means that external HTML files do not trigger project loading when they are the first and only file opened.

As a best-effort, we attempt to open a TS file with the same name, which will most often be the file for the component of the external template. This is not guaranteed to be the case, so the only way to guarantee this works correctly would be for developers to add the `**/*.html` files to the `includes` list of their `tsconfig[.lib].json`.

resolves #2149

### DIFF
--- a/server/src/session.ts
+++ b/server/src/session.ts
@@ -821,7 +821,20 @@ export class Session {
       // The content could be newer than that on disk. This could be due to
       // buffer in the user's editor which has not been saved to disk.
       // See https://github.com/angular/vscode-ng-language-service/issues/632
-      const result = this.projectService.openClientFile(filePath, text, scriptKind);
+      let result = this.projectService.openClientFile(filePath, text, scriptKind);
+      // If the first opened file is an HTML file and the project is a composite/solution-style project with references,
+      // TypeScript will _not_ open a project unless the file is explicitly included in the files/includes list.
+      // This is quite unlikely to be the case for HTML files. As a best-effort to fix this, we attempt to open
+      // a TS file with the same name. Most of the time, this is going to be the component file for the external template.
+      // https://github.com/angular/vscode-ng-language-service/issues/2149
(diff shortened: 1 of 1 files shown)


jarpoole (approved):
At least for my team, this fix would be 100% successful across our entire monorepo!

See https://github.com/angular/vscode-ng-language-service/issues/2149#issuecomment-2763979506

clydin (approved):

PR #2191, opened 2025-06-15
feat: add semantic tokens

JannikLassahn:
Add integration for semantic tokens provided by @angular/language-service.
Only supports  token type 'class' at the moment.


server/src/semantic_tokens.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as ts from 'typescript/lib/tsserverlibrary';
+import * as lsp from 'vscode-languageserver/node';
+
+export enum TokenEncodingConsts {
```
atscott, line 12: Would you mind linking to the TS definitions like you did in https://github.com/angular/angular/pull/60260? Context is pretty lost between the two sometimes

server/src/semantic_tokens.ts
```diff
@@ -0,0 +1,62 @@
...
+function getTokenTypeFromClassification(tsClassification: number): number|undefined {
+  if (tsClassification > TokenEncodingConsts.modifierMask) {
+    return (tsClassification >> TokenEncodingConsts.typeOffset) - 1;
+  }
+  return undefined;
+}
+
+function getTokenModifierFromClassification(tsClassification: number) {
+  return tsClassification & TokenEncodingConsts.modifierMask;
+}
```
atscott, line 62: Hmm, I wonder if this should really be something that the language service provides since it's so highly dependent on the offset and modifier mask actually matching what the language service returns.

server/src/semantic_tokens.ts
JannikLassahn: moved to language service

server/src/semantic_tokens.ts
JannikLassahn: I moved everything to `@angular/language-service/api` 
See PR https://github.com/angular/angular/pull/62251

atscott (approved):

atscott:
FYI, this was rolled back in https://github.com/angular/vscode-ng-language-service/pull/2197 due to https://github.com/angular/vscode-ng-language-service/issues/2196

JannikLassahn:
Sorry about that issue. I'll try to find a fix as soon as I have time!

PR #2192, opened 2025-06-20
feat(language-service): support to report the deprecated API in the t…

ivanwonder:
…emplate

In the Typescript Language Service, these diagnostics are reported as suggestion diagnostics. This will report the deprecated `Component`, `Directive`, etc.

### DIFF
--- a/server/src/utils.ts
+++ b/server/src/utils.ts
@@ -113,12 +113,14 @@ export function lspRangeToTsPositions(
  * @param relatedInfo
  */
 export function tsRelatedInformationToLspRelatedInformation(
-    scriptInfo: ts.server.ScriptInfo,
+    projectService: ts.server.ProjectService,
     relatedInfo?: ts.DiagnosticRelatedInformation[]): lsp.DiagnosticRelatedInformation[]|undefined {
   if (relatedInfo === undefined) return;
   const lspRelatedInfo: lsp.DiagnosticRelatedInformation[] = [];
   for (const info of relatedInfo) {
     if (info.file === undefined || info.start === undefined || info.length === undefined) continue;
+    const scriptInfo = projectService.getScriptInfo(info.file.fileName);
+    if (scriptInfo === undefined) continue;
     const textSpan: ts.TextSpan = {
       start: info.start,
       length: info.length,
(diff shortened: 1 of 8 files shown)


server/src/utils.ts
```diff
@@ -113,12 +113,14 @@ export function lspRangeToTsPositions(
...
 export function tsRelatedInformationToLspRelatedInformation(
-    scriptInfo: ts.server.ScriptInfo,
+    projectService: ts.server.ProjectService,
     relatedInfo?: ts.DiagnosticRelatedInformation[]): lsp.DiagnosticRelatedInformation[]|undefined {
   if (relatedInfo === undefined) return;
   const lspRelatedInfo: lsp.DiagnosticRelatedInformation[] = [];
   for (const info of relatedInfo) {
     if (info.file === undefined || info.start === undefined || info.length === undefined) continue;
+    const scriptInfo = projectService.getScriptInfo(info.file.fileName);
+    if (scriptInfo === undefined) continue;
```
ivanwonder, line 123: There is an issue here. The diagnostic file may differ from the one in 'info', so the range needs to be adjusted using the diagnostic file.

ivanwonder:
@atscott you can review this PR now.

server/src/utils.ts
atscott: I’m not quite following. The diagnostic isn’t already mapped to the appropriate info by this time? Should it be done in the language service?

ivanwonder: This issue is the transformation from the ts range to lsp range.

For example:

The diagnostic is located in the HTML, and there is a related diagnostic for the `ts`(Error occurs in the template of component ***). For the old logic, the script info is the HTML, but we should use the ts script info to transform the ts range. 

And we should not assume here that the related diagnostics all come from the same script info.

atscott: Right, the related info can be a different file. I’m not sure then that I follow this comment thread. I don’t see those edits here. Are you unsure if they’re right or saying there’s no way to fix it?

atscott: Got it! I understand now. Thanks for this.

atscott (approved):

ivanwonder:
> @ivanwonder I am unable to observe this working. Can you describe what this covers now that it's limited to components? Should it support inputs and outputs as well? Should it support inline templates and external templates or just one of the two?

Sorry, I have not updated the version of the language service. It supports the inline templates and external templates.

You can try again.

ivanwonder:
@atscott Just find a bug that will cause the extension to crash.

https://github.com/angular/angular/pull/63002 PTAL.

atscott:
> Sorry, I have not updated the version of the language service. It supports the inline templates and external templates.

Oh, sorry this was actually my mistake. I discovered the extension was broken recently and was waiting on a fix
https://github.com/angular/angular/pull/62811 upstream before testing. When that went in, I updated to 20.1.4 instead of -next to do the test!

atscott (approved):

